PR #6, opened 2024-08-27
Subscriptions

juligasa:
Proto definition for the new subscription API. Just Local subscriptions (no public subscriptions) to documents and spaces.


proto/activity/v1alpha/subscriptions.proto
```diff
@@ -0,0 +1,78 @@
...
+
+  // Remove a subscription.
+  rpc Unsubscribe(UnsubscribeRequest) returns (google.protobuf.Empty);
+
+  // Lists active subscriptions.
+  rpc ListSubscriptions(ListSubscriptionsRequest) returns (ListSubscriptionsResponse);
+}
+
+// Subscribe to a resource
+message SubscribeRequest {
```
ericvicenti, line 23: where is the path, @juligasa ?

proto/activity/v1alpha/subscriptions.proto
```diff
@@ -0,0 +1,78 @@
...
+    // The list of subscriptions.
+    repeated Subscription subscriptions = 1;
+
+    // The token to request the next page.
+    string next_page_token = 2;
+}
+
+
+// Description of the subscription item.
+message Subscription {
```
ericvicenti, line 58: I expect a path here too

ericvicenti (changes requested):

proto/activity/v1alpha/subscriptions.proto
juligasa: Oh, I maybe I should change `id` since it is ambiguous. Do you want an `url` field where you can input `hm://<account><path>` or you want two different fields `account`, `path` just like in `getDocument`? @ericvicenti

burdiyan: In all the new APIs we use account and path separately.

ericvicenti: yeah i would like to keep the APIs consistent

PR #21, opened 2024-09-10
Fix: Fallback to using IDs instead of saying 'Unititled Document'

burdiyan:
On the contacts page it says 'Untitled Document' (not 'Untitled Account' btw, for which I also have seen a function available). I changed to fall back to using IDs instead of this, because it's impossible to distinguish dozens of 'Untitled Documents' from one another.

Before:

Screenshot 2024-09-10 at 18 31 32

After:

Screenshot 2024-09-10 at 18 30 22

We could truncate the IDs, but they seem to be short-enough to be shown in full.


ericvicenti:
ok this is a pretty bad UX but hopefully we can solve it soon

PR #22, opened 2024-09-11
Buffered blocks on syncing

juligasa:
In order to save redundant fetches we should buffer the blobs


juligasa:
As we wait until the end to start indexing blobs, every block takes around 150ms to fetch and the connection times out. Next time discovery is called, we still don't have any blocks indexed so we start downloading them again, timing out. This did not happen before as we were indexing blocks as we downloaded them, so consecutive calls reduced the amount of wants and eventually it succeeded. We should index blobks in parallel as we download

PR #46, opened 2024-11-05
fix(frontend): fix blank page when daemon takes too long to start

burdiyan:
We had an unhandled case for the daemon state variable.

Fixes [SHM-72](https://linear.app/seedhypermedia/issue/SHM-72/blank-page-if-daemon-takes-long-to-start).


horacioh:
I'm reviewing this

PR #71, opened 2024-12-05
WIP: Library API definitions

burdiyan:
This PR has the API definitions necessary for the new Library home page. Please review @ericvicenti.


proto/documents/v3alpha/documents.proto
```diff
@@ -218,20 +371,57 @@ message DocumentListItem {
...
+  // ID of the most recent comment.
+  string latest_comment_id = 1;
+
+  // Timestamp of the most recent comment.
+  google.protobuf.Timestamp latest_comment_time = 2;
+
+  // Total number of comments.
+  int32 comment_count = 3;
+
+  // Timestamp of the most recent change.
```
ericvicenti, line 376: low pri: should maybe have the id of the latest change so we can look up what exactly changed

proto/documents/v3alpha/documents.proto
```diff
@@ -218,20 +371,57 @@ message DocumentListItem {
...
   string genesis = 8;
 
   // Current version of the document.
   string version = 9;
+
+  // Breadcrumbs of this document.
+  // The number of items in this list corresponds to the number of segments
+  // in the path of the document minus 1 (the document itself).
+  //
+  // E.g. for a document "/cars/jp/honda" there will be 2 breadcrumbs: "cars" and "jp".
```
ericvicenti, line 355: we also need the site name, not just the names of the parent docs

proto/documents/v3alpha/documents.proto
```diff
@@ -218,20 +371,57 @@ message DocumentListItem {
...
+  int32 comment_count = 3;
+
+  // Timestamp of the most recent change.
+  google.protobuf.Timestamp latest_change_time = 4;
+}
+
+// Basic information about ancestor documents in the path hierarchy.
+message Breadcrumb {
+  // Title of the document.
+  string title = 1;
```
ericvicenti, line 383: we should use "name" instead of "title" to be consistent with document metadata naming

proto/documents/v3alpha/documents.proto
```diff
@@ -23,8 +25,23 @@ service Documents {
...
+  // Lists all accounts.
+  //
+  // TODO(burdiyan): rename to spaces or sites.
+  rpc ListAccounts(ListAccountsRequest) returns (ListAccountsRequest);
+
+  // Lists documents in a directory of an account.
+  rpc ListDirectory(ListDirectoryRequest) returns (ListDirectoryResponse);
+
+  // Searches documents within our node.
+  rpc SearchDocuments(SearchDocumentsRequest) returns (SearchDocumentsResponse);
```
ericvicenti: rm search

proto/documents/v3alpha/documents.proto
```diff
@@ -110,6 +127,139 @@ message ListRootDocumentsResponse {
...
+
+  // Summary of the activity in the entire account.
+  ActivitySummary activity_summary = 3;
+
+  // Whether the user is subscribed to this account.
+  bool is_subscribed = 4;
+}
+
+// Request to list a directory.
+message ListDirectryRequest {
```
ericvicenti, line 181: needs subscribed_only

proto/documents/v3alpha/documents.proto
```diff
@@ -110,6 +127,139 @@ message ListRootDocumentsResponse {
...
+
+  // Optional. Pagination token.
+  string page_token = 2;
+
+  // Required. ID of the account to list the directory for.
+  string account = 3;
+
+  // Required. Path of the directory to list documents for.
+  // Empty string means root directory.
+  // The root document is considered a part of the root directory.
```
ericvicenti, line 193: >The root document is considered a part of the root directory.

I'm not sure about this, we always throw away the root doc when it appears in our lists

ericvicenti (changes requested):
some initial thoughts

can't wait to do this all over again when we have graphql

proto/documents/v3alpha/documents.proto
burdiyan: I thought we'd be already having the site name in memory when calling methods for listing documents?

proto/documents/v3alpha/documents.proto
burdiyan: I thought about that too, but wasn't sure. I was actually surprised to see that we ended up with `name` instead of `title` :)

proto/documents/v3alpha/documents.proto
burdiyan: I did.

proto/documents/v3alpha/documents.proto
burdiyan: Right. Missed that one.

burdiyan: Also a typo in there.

proto/documents/v3alpha/documents.proto
burdiyan: Hm... OK. Won't put it there then.

proto/documents/v3alpha/documents.proto
burdiyan: DONE.

proto/documents/v3alpha/documents.proto
burdiyan: Done.

proto/documents/v3alpha/documents.proto
burdiyan: I removed this comment. Although this sounds a bit contradictory to what you said here, I guess? https://github.com/seed-hypermedia/seed/pull/71#discussion_r1871751811

proto/documents/v3alpha/documents.proto
ericvicenti: I can get the change details by looking directly at the CID data

And I would use it right away. Most important actually is to know **who** made the most recent change (plus the time)

proto/documents/v3alpha/documents.proto
burdiyan: I'll add the change ID

proto/documents/v3alpha/documents.proto
```diff
@@ -218,20 +334,57 @@ message DocumentListItem {
...
+  // If we don't have a document for some of the path segements for whatever reason,
+  // the corresponding breadcrumb segment will be null.
+  repeated Breadcrumb breadcrumbs = 10;
+
+  // Summary of the activity on the document.
+  ActivitySummary activity_summary = 11;
+}
+
+// Summary of the activity on a document or account.
+message ActivitySummary {
```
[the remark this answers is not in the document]
ericvicenti: for now that should be fine! although delete would be nice-to-have as well

PR #72, opened 2024-12-11
Desktop Dev Builds

ericvicenti:
New Dev Builds Release Process by @horacioh 

Dev Builds are:
- [x] Are released on every push to main branch
- Publishes to its own auto-update channel. (when you are using dev, you will be prompted to update very frequently!)
  - [x] Delivered through S3 bucket instead of GH, to avoid spamming release page and because electron forge requires it
- [x] Use a different DB path (`Seed-dev`)
- [x] Use Hypermedia testnet to avoid polluting mainnet with experimental and junk content
- [x] Has a different app name - "Seed Dev"
- [x] Has a white app icon


ericvicenti:
I know this is still a WIP but I can't wait to merge it!

PR #74, opened 2025-01-16
Implement nested document attributes

burdiyan:


horacioh:
@burdiyan I merged the outline property change to this branch but the metadata of the document is coming empty.

do I need to use a separate method to update the metadata apart from `setMetadata`?

burdiyan:
> @burdiyan I merged the outline property change to this branch but the metadata of the document is coming empty.
> 
> do I need to use a separate method to update the metadata apart from `setMetadata`?

@horacioh you should be using the new `setAttribute` operation instead of `setMetadata`, especially when values are not simple strings (like in the `showOutline` case).

The old `setMetadata` still works though, and should come back empty. I have lots of tests that still use it.

Tell me exactly what you are doing so I could reproduce it.

horacioh:
@burdiyan can I replace `setMetadata` with `setAttribute` then?

horacioh:
@burdiyan fixed. now it worked

horacioh (approved):

burdiyan:
> @burdiyan can I replace `setMetadata` with `setAttribute` then?

Yes, you can. And you probably should.

burdiyan:
@horacioh When I run the code in this branch my document list has all documents untitled.

image

Notices that the `metadata` and `attributes` have changed from being simple `map<string, string>` to `google.protobuf.Struct`. So probably many other parts of the code that touch those part need to change.

PR #79, opened 2025-02-10
feat: editor on web pt1

horacioh:
- create a new package: @shm/editor
- fix dev env between @shm/ui and @shm/shared when running apps
- we can now run @shm/web on dev mode for some stuff.
- we can comment on the web (just block hierarchy and paragraphs


ericvicenti:
Great, lets merge and keep on iterating!

PR #81, opened 2025-02-13
Feat/content width

ericvicenti:
Enables the content width feature, allowing users to bring back the width of newspaper layout


horacioh (approved):
overall LGTM. there are some minore details needed to be fixed but the overall functionality of the new property works.

PR #85, opened 2025-03-04
feat: Onboarding part 1

horacioh:


frontend/apps/desktop/src/models/daemon.ts
```diff
@@ -111,6 +108,33 @@ export function useMyAccounts() {
...
+/**
+ * Returns a list of keys from the daemon.
+ * This is a hook that is used to list keys from the daemon.
+ * It is used to check if the user has any accounts.
+ * If the user has no accounts, it will show the onboarding screen.
+ * If the user has accounts, it will show the main app.
+ *
+ * @returns
+ */
+export function useListKeys() {
```
ericvicenti, line 120: whats wrong with useMyAccountIds ? Why not use react query?

frontend/apps/desktop/src/utils/navigation-container.tsx
```diff
@@ -48,6 +48,12 @@ export function NavigationContainer({
     return navigation.state.subscribe(() => {
       const state = navigation.state.get()
       send('windowNavState', state)
+      // Emit route change event to main process
+      send('routeChange', {
```
ericvicenti, line 52: I don't see why this is necessary. The middle end already has the navigation state at this point, sent on line 50. So why do we send all this stuff again in a separate event?

frontend/apps/desktop/src/find-in-page.tsx
```diff
@@ -1,22 +1,26 @@
...
 import '@tamagui/font-inter/css/700.css'
-import React from 'react'
+import React, {useEffect, useState} from 'react'
 import ReactDOM from 'react-dom/client'
 import {FindInPage} from './pages/find-in-page'
 
 function FindInPageView() {
-  // @ts-expect-error
-  const darkMode = useStream<boolean>(window.darkMode)
+  const [darkMode, setDarkMode] = useState(false)
```
ericvicenti, line 10: it looks like darkMode will always start off as false? This is the reason we have useSyncExternalStore inside useStream

And, what is wrong with useStream?

frontend/apps/desktop/src/app-onboarding-store.ts
```diff
@@ -0,0 +1,80 @@
...
+const store = new Store<OnboardingState>({
+  name: 'onboarding',
+  cwd: path.join(app?.getPath('userData') || process.cwd(), DESKTOP_APPDATA),
+  defaults: ONBOARDING_INITIAL_STATE,
+})
+
+const getInitialState = (): OnboardingState => ONBOARDING_INITIAL_STATE
+
+export function setupOnboardingHandlers() {
+  ipcMain.on('get-onboarding-state', (event) => {
```
ericvicenti, line 28: wouldn't it be easier to use TRPC like we do for everything else?

Like recents, favorites, settings, the "host" store stuff in my other branch, gatewaySettings, recentSigners, drafts, commentDrafts.

I don't understand why we are breaking convention here by using ipc directly instead of TRPC+invalidation. It seems very inconvenient and inconsistent in this codebase.

frontend/apps/desktop/src/window-manager.ts
ericvicenti, line 1: what changed in this file? Its very hard to review the code when files get moved without `git mv`

frontend/apps/desktop/src/components/publish-site.tsx
ericvicenti, line 1: uh oh! it looks like this was rebased on my branch?! This is a broken experience in this state, so we can't merge it

ericvicenti (changes requested):
I see some bits of your secret project in here, exciting! 😁

horacioh:
This was the first attempt. But I removed them in favor of the real one

ericvicenti:
What do you mean removed? I see lots of stuff about metrics and front-analytics in this PR

frontend/apps/desktop/src/models/daemon.ts
horacioh: because I want as minimal things as possible in the main process. I don't want to deal with react query stuff in there

frontend/apps/desktop/src/utils/navigation-container.tsx
horacioh: I can remove it if you are confident this is being sent before.

frontend/apps/desktop/src/find-in-page.tsx
horacioh: is this important for this PR to be merged?

frontend/apps/desktop/src/app-onboarding-store.ts
horacioh: like I said above, I want the minimal things as possible in the middle end. and when I was using TRPC I was getting flashes on the first load of the app

frontend/apps/desktop/src/window-manager.ts
horacioh: we had app-windows. and I changed it to window-manager and there are better structure on how we load stuff following the proper Electron conventions.

frontend/apps/desktop/src/components/publish-site.tsx
horacioh: this is rebased with main. not with any other branch. I dunno which branch are you talking about

frontend/apps/desktop/src/models/daemon.ts
ericvicenti: oh! Wait yeah this is the main process. So why does it have React useEffect? How does this even work?

frontend/apps/desktop/src/window-manager.ts
ericvicenti: Is it possible to use git mv so I can review the changes?

frontend/apps/desktop/src/components/publish-site.tsx
ericvicenti: This file has code from `feat/hosting-flow`, not the main branch

frontend/apps/desktop/src/find-in-page.tsx
ericvicenti: No I'm happy to merge anyways.

But I'm trying to understand why you moved to useState+useEffect because this is not the right approach in my opinion. useSyncExternalStore is meant for this, and that is how useStream works. And you deleted useStream here. Why?

frontend/apps/desktop/src/utils/navigation-container.tsx
ericvicenti: yes I'm confident because the whole state is sent right here on the previous line `send('windowNavState', state)`

frontend/apps/desktop/src/components/publish-site.tsx
horacioh: I dunno how that happened. I did not do anything with that branch

frontend/apps/desktop/src/window-manager.ts
horacioh: can we review this after merge? is it super important to review this now?

frontend/apps/desktop/src/models/daemon.ts
ericvicenti: what do you mean, in the main process? This is being used in the frontend. We have already loaded react-query at this time. Can we please just `useMyAccountIds` like we normally do?

frontend/apps/desktop/src/components/publish-site.tsx
ericvicenti: yeah my mistake! For a few minutes I had accidentally merged some changes to main. And thats when you rebased 😅

PR #97, opened 2025-04-22
Feat/full text search

juligasa:
Full text search based on [Sqlite plugin](https://sqlite.org/fts5.html)


backend/storage/schema.sql
```diff
@@ -229,3 +229,18 @@ CREATE TABLE wallets (
...
 CREATE INDEX wallets_by_account ON wallets (account);
+
+-- Stores text content to to a full text search
+-- https://sqlite.org/fts5.html.
+
+CREATE VIRTUAL TABLE fts USING fts5(
+    -- The text content to be indexed.
+    raw_content, 
+    -- The type of the content being indexed.
+    type,
```
burdiyan, line 240: @juligasa What does type mean here? Can you document what possible types are supported here?

backend/storage/schema.sql
```diff
@@ -229,3 +229,18 @@ CREATE TABLE wallets (
...
+-- Stores text content to to a full text search
+-- https://sqlite.org/fts5.html.
+
+CREATE VIRTUAL TABLE fts USING fts5(
+    -- The text content to be indexed.
+    raw_content, 
+    -- The type of the content being indexed.
+    type, 
+    -- The multihash of the blob containting the content.
+    multihash,
```
burdiyan, line 242: @juligasa few questions:

1. Is there a way to have indexes on columns of FTS5 tables?
2. Any particular reason we use multihash here instead of the internal integer ID, or full CID?

One of the concerns I have is that we might need to clean up the FTS5 table after we've deleted a blob (which we don't currently do, but I bet implementing GC is not too far away).

burdiyan: Actually, I just remembered that it was one of the reasons I wanted to use FTS5 with external content table: a document ID in FTS5 can only be a rowid (IIRC), and our blobs table have rowids already. So some of the cleanup bookkeeping could be easier to implement this way.

juligasa: I changed it so that we store the blobID instead. The idea is doing an incremental inplementation, First a quick full end to end full text search and then optimizations.

burdiyan:
@juligasa I see there're some linter errors. Just FYI, you can run the linter locally only for the code you've touched (same thing the CI would check) by running this command:

```
golangci-lint run --new-from-rev=$(git merge-base --fork-point origin/main) ./backend/...
```

Make sure to install a more or less recent version of `golangci-lint` (it's not in our nix environment due to several obscure problems).

backend/api/entities/v1alpha/entities.go
```diff
@@ -206,36 +248,96 @@ func (srv *Server) SearchEntities(ctx context.Context, in *entities.SearchEntiti
...
-	for i, match := range matches {
-		if match.Score < 0 {
-			limit++
-			continue
-		}
-		if i >= limit {
-			break
+	var numTitles = len(contents)
+	var bodyMatches []fuzzy.Match
+	cleanQuery := strings.ReplaceAll(strings.ReplaceAll(strings.ReplaceAll(strings.ReplaceAll(in.Query, ",", ""), "$", ""), "^", ""), "*", "")
```
burdiyan, line 282: @juligasa can you please write some comments about what this is doing? I find it pretty confusing at first sight, and would appreciate some details.

PR #109, opened 2025-06-30
Email notifications for document mentions and document changes

iskaktoltay:
- Added email notifications for changes of an owned document
- Added email notifications for mentions inside of a document

### DIFF
--- a/frontend/apps/emails/components/EmailContent.tsx
+++ b/frontend/apps/emails/components/EmailContent.tsx
@@ -6,7 +6,7 @@ import {
   MjmlSection,
   MjmlText,
 } from '@faire/mjml-react'
-import {BlockNode} from '@shm/shared'
+import {BlockNode, createWebHMUrl, unpackHmId} from '@shm/shared'
 import {format} from 'date-fns'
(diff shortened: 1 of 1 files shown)


frontend/apps/emails/components/EmailContent.tsx
```diff
@@ -353,10 +382,27 @@ function renderInlineTextWithAnnotations(
...
+          hostname: unpacked.hostname ?? null,
+        })
+      }
+      annotatedText = `<a href="${href}" style="color: #346DB7;">${annotatedText}</a>`
     } else if (annotation.type === 'Embed') {
       const resolved = resolvedNames?.[annotation.link] || annotation.link
-      annotatedText = `<a href="${annotation.link}" style="color: #008060;">@${resolved}</a>`
+
+      let href = annotation.link
+      if (href?.startsWith('hm://')) {
```
ericvicenti, line 398: you can just try unpacking it right away. then just handle the case where the unpacked ID is null

ericvicenti (approved):
Looks good, thanks!

PR #112, opened 2025-07-16
Improve FTS performance

juligasa:
Now we precompute the most expensive data (updating the latest version where a match has been seen) when indexing. 
Indexing overhead is unnoticeable from before due to a creation of a new indexed table. This was necessary because the fts table is a virtual table and we could not create the necessary indexes to speed up the query, leading to large reindexing times. 
Precomputing the information needed unburden the search engine with excessive computation work, which should alleviate server requirements

### DIFF
--- a/backend/storage/storage_migrations.go
+++ b/backend/storage/storage_migrations.go
@@ -57,6 +57,42 @@ type migration struct {
 //
 // In case of even the most minor doubts, consult with the team before adding a new migration, and submit the code to review if needed.
 var migrations = []migration{
+	{Version: "2025-07-15.01", Run: func(_ *Store, conn *sqlite.Conn) error {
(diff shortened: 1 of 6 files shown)


backend/storage/storage_migrations.go
```diff
@@ -57,6 +57,45 @@ type migration struct {
 //
 // In case of even the most minor doubts, consult with the team before adding a new migration, and submit the code to review if needed.
 var migrations = []migration{
+	{Version: "2025-07-15.01", Run: func(_ *Store, conn *sqlite.Conn) error {
```
burdiyan, line 60: @juligasa ExecScript lets you write multiple statements and execute them in one go (must have the terminating semicolons). Not that it affects the functionality, but it makes it a bit easier to read I think.

PR #115, opened 2025-08-11
Feat/types fixed

horacioh:
This PR suppresses all current typescript errors in the codebase. this will help us to actually use Typescript to prevent any new type errors in the codebase, while letting us fix the current errors commented via directives.

I do have another commit that will add a check on every push and PR for new directives. If any new directive is found, it will fail.

the next step is to create a simple dashboard that tracks all the directives in the codebase to help the typescript team fix them incrementally.


horacioh:
I merged this because it will be harder to merge if we keep it unmerged. and is actually not hurting us, but the opposite.

PR #125, opened 2025-11-04
WIP: Server Push

burdiyan:


backend/api/daemon/v1alpha/daemon.go
```diff
@@ -75,6 +77,58 @@ func (srv *Server) RegisterServer(rpc grpc.ServiceRegistrar) {
 	daemon.RegisterDaemonServer(rpc, srv)
 }
 
+// SyncResourcesWithPeer implements the corresponding gRPC method.
+func (srv *Server) SyncResourcesWithPeer(req *daemon.SyncResourcesWithPeerRequest, stream grpc.ServerStreamingServer[daemon.SyncingProgress]) error {
```
burdiyan: @juligasa I'm not sure using streaming here is a good idea. If the client disconnects we probably don't want to cancel the syncing that was initiated, so it probably shouldn't be scoped to the API handler lifetime. WDYT?

backend/hmnet/syncing/syncing.go
```diff
@@ -448,6 +449,54 @@ func (s *Service) syncWithManyPeers(ctx context.Context, subsMap subscriptionMap
 	return res
 }
 
+// SyncResourcesWithPeer syncs the given resources with a specific peer.
+// method is exposed for use externally, (pushing content to a peer)
+func (s *Service) SyncResourcesWithPeer(ctx context.Context, pid peer.ID, resources []string) error {
+	dkeys := make(colx.HashSet[discoveryKey], len(resources))
+	rkeys := map[string]bool{}
+	var hmRe = regexp.MustCompile(
```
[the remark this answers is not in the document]
juligasa: Ok, I removed unused groups and put it as a global variable. Once compiled the rest is quick so I would like to keep it like that so that we can accept more resources in the near future

backend/api/daemon/v1alpha/daemon.go
juligasa: I think you are absolutely right in not attaching the the syncing lifecycle to the caller context. I just wanted to set a reasonable timeout and I did not know which one so I got the one from the call.
As for the usefulness of the stream I think it is useful. You can totally skip the streamed result and just look at the error return if you want. But frontend team has been complaining in the past that they wanted some kind of live feedback and this was quite cheap to add, so I imagine our new publish functionality like a progress bar on the bottom that eventually completes.

backend/api/daemon/v1alpha/daemon.go
```diff
@@ -75,6 +77,58 @@ func (srv *Server) RegisterServer(rpc grpc.ServiceRegistrar) {
...
+				PeersSyncedOk:   prog.PeersSyncedOK.Load(),
+				PeersFailed:     prog.PeersFailed.Load(),
+				BlobsDiscovered: prog.BlobsDiscovered.Load(),
+				BlobsDownloaded: prog.BlobsDownloaded.Load(),
+				BlobsFailed:     prog.BlobsFailed.Load(),
+			}
+			_ = stream.Send(out) // ignore send error on termination
+			return syncErr
+		case <-prog.Updates():
+			out := &daemon.SyncingProgress{
```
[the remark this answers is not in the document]
juligasa: huh? We do reuse it. The one you pointed out is the protobuf equivalent of it.

backend/hmnet/syncing/discovery_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"seed/backend/util/sqlite/sqlitex"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestLoadExternalStore(t *testing.T) {
+	t.Skip("This uses external database for local test only")
+	t.Parallel()
+	pool := loadLocalDB(t, "/home/julio/.config/Seed-local/daemon/db/db.sqlite")
```
burdiyan: @juligasa I guess this test doesn't need to be here now?

BTW, do we have any proper tests for the new push functionality?

backend/api/documents/v3alpha/resources.go
```diff
@@ -2,23 +2,137 @@ package documents
...
 	cbornode "github.com/ipfs/go-ipld-cbor"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
 
+// PushResourcesToPeer implements the corresponding gRPC method.
+func (srv *Server) PushResourcesToPeer(req *documents.PushResourcesToPeerRequest, streamTx grpc.ServerStreamingServer[documents.SyncingProgress]) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
burdiyan: Do we need this strict timeout here? I'm afraid sometimes 30 seconds may be too low. What if the connection is slow, or there's a lot of data to sync?

backend/daemon/daemon_e2e_test.go
```diff
@@ -842,17 +842,354 @@ func TestSubscriptions(t *testing.T) {
...
 
-	docGotten, err := bob.RPC.DocumentsV3.GetDocument(ctx, &documents.GetDocumentRequest{
+	require.Eventually(t, func() bool {
+		docGotten, err := bob.RPC.DocumentsV3.GetDocument(ctx, &documents.GetDocumentRequest{
+			Account: aliceHome.Account,
+			Path:    aliceHome.Path,
+		})
+		require.NoError(t, err)
+		return len(aliceHome.Content) == len(docGotten.Content)
+	}, time.Second*2, time.Millisecond*100, "We should have three comments, including carol's")
```
burdiyan: @juligasa is this error message accurate? We query bob here for Alice's document, then we check that we receive the same version, while the error message is talking about three *comments* and says something about Carol. Is this correct?

PR #127, opened 2025-12-01
Feat/api router

ericvicenti:
https://seedteamtalks.hyper.media/tech-talks/query-router-proposal?v=bafy2bzacect5jcnf6qbxsqmtdgqghr5cikvfhogx3jfzvpbvw2e7lyy3bwntg&l

### DIFF
--- a/frontend/apps/desktop/src/pages/document.tsx
+++ b/frontend/apps/desktop/src/pages/document.tsx
@@ -635,10 +635,7 @@ function DocPageHeader({
   documentTools?: any
 }) {
   const authors = useMemo(() => document?.authors || [], [document])
-  const authorIds = useMemo(
-    () => authors?.map((a) => hmId(a)) || [],
-    [authors],
-  )
+  const authorIds = useMemo(() => authors?.map((a) => hmId(a)) || [], [authors])
   const authorResources = useResources(authorIds, {subscribed: true})
   const authorContacts = useContactsMetadata(authors || [])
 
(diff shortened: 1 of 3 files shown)


ericvicenti:
I was asking for help to see if there are any regressions in this code. We can do some of these fixes in the main branch because they seem unrelated to the query router and the issues are pre-existing.

## 1. Add Recursion Guard to BatchAccounts
We should never have recursive accounts. Good idea to guard against though. Question: why only protect within BatchGetAccounts, shouldn't we guard against recursion in resource and account query as well?

This is not something we had before, so I think we can do it later

## 2. Validation logic
Hm, the error would be slightly more clear but thats the only difference. The error call stack would be specific enough I think

## 3. Event error logging
This seems unrelated to this PR. The same issue existed before, and we can fix this later if it is important

## 4. Tests
Good idea! But this wasn't tested before so I would recommend we do this later in the main branch. We should use a fixture database so that the tests for loading don't have a complicated setup. The test needs to run a daemon and the web site, so things get a little complicated

## 5. Type guard file
I can't figure out how to avoid this 'any' but the important thing is that the code inside the API routes is type-safe, and the code outside is type-safe. So I think this PR is still an overall improvement.

## 6. This is entirely unrelated to the pull request

## 7. We have specific error types already

Search for `HMRedirectError` and `HMNotFoundError` for the underlying `Error` subclasses

And for the data that the endpoints return to indicate errors: search for `HMResourceNotFound` and `HMResourceTombstone` which means 'deleted'

## Documentation
Actually the migration guide should be deleted because the migration is complete. Of course AI likes it because it was written by AI 😂. I'll remove it.

## Conclusion
@horacioh do you think we are ready to merge?

horacioh:
well if we can find a way to remove the `as any` that would be great. because this PR introduces more code than what is removing, so if we can at least make sure types are correct here that would be awesome!

I understand the recursive guard was not there before. That is not a justification not to add it here (with tests). We are trying to improve the codebase overall!

So IMHO, if we can do these two things is ready to merge

PR #394, opened 2026-03-24
perf(search): optimize autocomplete and add instrumentation

### LINKED ISSUE #391: Bug: autocomplete inside the editor is slow
## Problem

after typing `@` inside the editor, searching for documents is slow. search in the omnibar is also slow. I want to know why search is slow.

horacioh:
## Summary

- **Backend**: Batch pre-fetch deletion status for comments (eliminates N+1 queries), trim results before expensive post-processing, add comprehensive timing instrumentation throughout search pipeline
- **Frontend**: Replace `useDeferredValue` with explicit 200ms debounce, add pageSize parameter, optimize result sorting (contacts/latest first), add performance logging across search stack
- **Tests**: Fix circular dependency in mocks by avoiding `vi.importActual` in navigation mock
- **Search strategy**: Use keyword-only search for queries <3 chars, hybrid for longer queries

## Changes

**Backend optimizations:**
- New `qBatchDeletedComments` query replaces N+1 deletion checks with single batch query
- Results trimmed to 2x page size before post-processing to avoid expensive per-result operations
- Debug logging at each stage (search, enrichment, dedup, authority ranking) with timing metrics

**Frontend improvements:**
- Debounce search input with explicit 200ms delay instead of `useDeferredValue`
- Pass `pageSize: 50` to API, enabling backend result trimming
- Search type selection based on query length (keyword-only for short queries)
- Better result sorting: contacts first, latest versions first, preserve backend scores
- Performance instrumentation in search hook, API layer, and inline mentions

**Test fixes:**
- Resolve circular dependency: `navigation → routing → utils barrel → url-to-route → navigation` by providing manual mocks instead of using `vi.importActual`

Fixes #391

### DIFF
--- a/backend/api/entities/v1alpha/entities.go
+++ b/backend/api/entities/v1alpha/entities.go
@@ -996,11 +996,6 @@ func (srv *Server) SearchEntities(ctx context.Context, in *entpb.SearchEntitiesR
 	}); err != nil {
 		return nil, err
 	}
-	srv.log.Debug("SearchEnrichmentCompleted",
-		zap.String("query", cleanQuery),
-		zap.Duration("enrichDur", time.Since(enrichStart)),
-		zap.Int("rawResults", len(searchResults)),
-	)
 	seen := make(map[string]int)
 	var uniqueResults []fullDataSearchResult
 	var uniqueBodyMatches []fuzzy.Match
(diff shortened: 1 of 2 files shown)


horacioh:
@juligasa the intention is not to leave the traces. I can remove them!
Thanks for the review

PR #450, opened 2026-04-13
perf(desktop): fix performance regression since v2026.2.6

horacioh:
## Summary

Reduce embeddings indexer resource consumption to address desktop performance regression since v2026.2.6. The frontend polling/subscription fixes were shipped in #444 — this PR covers the remaining backend tuning.

- Reduce embedding index pass size from 10 to 5 (less memory per pass)
- Increase sleep between passes from 500ms to 2s (more CPU yielding)
- Increase embedding run interval from 1min to 5min (less frequent indexing)

These changes significantly reduce CPU and memory pressure from the background embeddings indexer, especially on low-RAM/CPU machines.

## Test plan

- [ ] Open app with embeddings enabled, monitor CPU usage at idle — should stay under 30%
- [ ] Verify embedding indexer still completes passes (check daemon logs for indexing activity)
- [ ] Confirm no regression in semantic search quality (slower indexing, same results)

### DIFF
--- a/backend/llm/embedding.go
+++ b/backend/llm/embedding.go
@@ -25,13 +25,13 @@ const (
 	// DefaultEmbeddingIndexPassSize is the default number of FTS rows to keep in memory per pass.
 	// After each pass, the embedder sleeps for a short time to avoid starving the CPU.
 	// Adjust the sleep duration via WithSleepPerPass.
-	DefaultEmbeddingIndexPassSize = 10
+	DefaultEmbeddingIndexPassSize = 5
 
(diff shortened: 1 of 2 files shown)


juligasa:
The embeddings throttling is not the problem, i think they are ok the way they are right now. We solved the most destructuve performance problems here https://github.com/seed-hypermedia/seed/commit/c11b4b7281f4d6826da0546baca84d5419ce1127 and the more fine grained issues are bein implemented here https://github.com/seed-hypermedia/seed/tree/fix/sync-scheduler-priorities will create a PR soon, so nothing to do here IMO

PR #451, opened 2026-04-14
syncing scheduler priorities

juligasa:
We should not allow scheduled syncing (sites we joined, subscriptions, etc) to interfere with what the user wants in real time. We should identify hot task to take preccedence over backgorund syncing
Also there are some performance side issues with the way we sync peers causing slow down of the important syncing path

### DIFF
--- a/backend/hmnet/syncing/scheduler.go
+++ b/backend/hmnet/syncing/scheduler.go
@@ -16,13 +16,7 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-// Task timing defaults. Installed into every new scheduler and may be
-// overridden per-instance before run() is called (tests exercise heartbeat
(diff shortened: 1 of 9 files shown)


backend/hmnet/syncing/scheduler.go
```diff
@@ -51,11 +56,16 @@ type taskHandle struct {
 
 	// Scheduling fields.
 	queueIndex   maybe.Value[int]
+	queueTier    uint8 // tierHot or tierCold. Snapshot of the task's tier at last enqueue; may go stale.
```
burdiyan: > Snapshot of the task's tier at last enqueue; may go stale.
What does this phrase mean? @juligasa

juligasa: simplified the logic now

PR #458, opened 2026-04-14
feat(publish): add optional message field to Ref blobs across the stack

horacioh:
publish message

## Summary

- Adds an optional `message` field to `Ref` blobs (proto, Go blob model, and CBOR encoding), analogous to a git commit message
- Surfaces the `message` field through `CreateDocumentChangeRequest`, `CreateRefRequest`, and the `Ref` response proto
- Threads the message through the full backend stack: `blob.NewRef`, `blob.NewRefRedirect`, `docmodel.Document.Ref`, and all indexing paths
- Exposes `message` in the frontend client types (`PublishDocumentInput`, `CreateVersionRefInput`, `CreateTombstoneRefInput`, `CreateRedirectRefInput`, `SignDocumentChangeInput`)
- Adds a publish message textarea to the desktop `PublishDraftButton` UI
- Adds `-m, --message` CLI flag to both `publish` and `update` document commands
- Fetches and displays the `message` field in the activity feed's `doc-update` events
- Updates generated protobuf files for both Go and TypeScript


ericvicenti:
I believe we should be adding fields to the Change blob, not the Ref blob, in the case of commit messages.

Also, PLEASE lets allow the API caller to append arbitrary metadata, which can be used for many things in addition to the Change blob. (Specifically I want to add metadata about the Agent and Context which motivated the Change)

ericvicenti:
cc @burdiyan

burdiyan:
I agree with Eric. I think descriptions should be added on Changes not on Refs. Just like in Git description is attached to Commits, not on Refs.

Refs in our system are meant to be garbage-collectable, and we don't even expose them in the UI anywhere as thing.

horacioh:
update and wrote a document with the new decision: https://seedteamtalks.hyper.media/projects/optional-publish-messages/publish-messages-implementation-decisions

horacioh:
@burdiyan I did change this to save the message in the change. please review

docs/publish-message.md
```diff
@@ -0,0 +1,50 @@
...
+
+- A Change is the unit of "what changed" in a document — analogous to a git commit.
+- A Ref is a pointer/branch; tombstones and redirects produce only Refs and have no content delta to describe.
+- Aligning the message with the commit-equivalent (Change) keeps the meaning consistent: only content publishes carry one.
+
+The field uses `omitempty` so existing Changes without a message remain valid. No migration is needed.
+
+### Stack
+
+- **Proto**: `string message` field on `CreateDocumentChangeRequest`, `PrepareChangeRequest`, and `DocumentChangeInfo`. `Ref.message` and `CreateRefRequest.message` removed (field numbers reserved).
```
burdiyan: This line assumes there was ever a message on the Ref, but it was only in this branch and was never released. If we merge these markdown docs into main it could become confusing, in a few places this assumes some prior context that's not true anymore.

@horacioh

proto/documents/v3alpha/documents.proto
```diff
@@ -227,6 +230,10 @@ message PrepareChangeRequest {
   // Can only be specified here when creating the document for the first time,
   // i.e. when `base_version` is empty.
   ResourceVisibility visibility = 6;
```
burdiyan: I know this has been there before, but it seems irrelevant, because visibility is specified on Refs not on Changes. We should probably remove this, because the client that uses this API is responsible for signing the correct Ref.

proto/documents/v3alpha/documents.proto
```diff
@@ -651,6 +658,11 @@ message CreateRefRequest {
 
   // Output only. Visibility of the document.
   ResourceVisibility visibility = 8;
+
+  // Field 9 was previously used for a Ref-level message. Messages now live on
+  // the Change blob (see CreateDocumentChangeRequest.message). Reserved to
+  // prevent accidental reuse.
+  reserved 9;
```
burdiyan: Ref messages were never released. We shouldn't reserve the field that was never released.

proto/documents/v3alpha/documents.proto
```diff
@@ -1004,6 +1020,11 @@ message Ref {
 
   // Information about the generation of the Ref.
   GenerationInfo generation_info = 8;
+
+  // Field 9 was previously used for a Ref-level message. Messages now live on
+  // the Change blob (see DocumentChangeInfo.message). Reserved to prevent
+  // accidental reuse.
+  reserved 9;
```
burdiyan: Same as https://github.com/seed-hypermedia/seed/pull/458/changes#r3240004681.

docs/publish-message-decisions.md
```diff
@@ -0,0 +1,94 @@
...
+For SQL-side queryability, the indexer copies `Message` into `structural_blobs.extra_attrs->>'message'`, alongside `title` and `metadata`. This lets the API surface the message without re-decoding the blob.
+
+## API surface
+
+Proto schema changes (`proto/documents/v3alpha/documents.proto`):
+
+- **Added** `string message = 9;` on `CreateDocumentChangeRequest` (kept from the original Ref-message branch — same field number, different semantics).
+- **Added** `string message = 7;` on `PrepareChangeRequest` so the daemon can embed the message into the unsigned Change blob it returns to client-side signers.
+- **Added** `string message = 5;` on `DocumentChangeInfo` so `GetDocumentChange` and `ListDocumentChanges` surface it.
+- **Removed** `string message = 9;` from `CreateRefRequest`. Field number reserved (`reserved 9;`).
```
burdiyan: Another example of some stale context being reflected in the final document.

burdiyan (commented):
The backend changes look good. I haven't reviewed the frontend.

Not sure about the relevance of those markdown documents — at the very least they should probably not carry assumptions about the message field being previously on the Refs, because this was never released.

horacioh:
@burdiyan I did all the recommended changes. let me know if you can check this

PR #461, opened 2026-04-14
fix(web): open the profile view when "My Profile" is clicked

horacioh:
The account menu used `key: 'profile'` to navigate, but on web the
profile view is rendered for `key: 'site-profile'`. Even though the
URL produced by both was equivalent, the in-memory NavContext kept
the wrong key after internal navigation, so `ResourcePage` fell
through to render the home document instead of the profile.

Use `key: 'site-profile'` with `tab: 'profile'` (matching the rest
of the web app) so both the mobile sheet and the desktop dropdown
in the account footer open the account's profile.

### DIFF
--- a/frontend/apps/web/app/web-utils.tsx
+++ b/frontend/apps/web/app/web-utils.tsx
@@ -153,7 +153,7 @@ export function WebAccountFooter({
         className="hover:bg-accent flex w-full items-center gap-3 px-4 py-3 text-left"
         onClick={() => {
           if (accountId) {
-            navigate({key: 'profile', id: hmId(accountId, {latest: true})})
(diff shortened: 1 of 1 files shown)


horacioh:
@ericvicenti why web has a different key for profiles than the desktop app?

ericvicenti:
Maybe we have a bug here, but to provide some context: on desktop app, the profile exists as a standalone page, and it should show activity for all sites. On web, it only exists within the same site, and the activity is meant to be filtered to that specific site.

So I guess that is the main explanation for there being different keys

ericvicenti:
On desktop the site-specific profiles also exist. But when you navigate to the profile it will take you to the generic one that exists outside the site

PR #464, opened 2026-04-16
upgrade libp2p , ipfs and go version

juligasa:
We have had some performance issues where we spawn multiple mini routines, and upgrading library versions and go versions may help.
The latest version of Go introduces a new garbage collector by default. This garbage collector seem to have been optimized for lots of tiny allocations, which is exactly our case.
It also introduces some improvements in CGO, which we also use a lot with SQLite and llama.cpp integration. Supposedly all CGO calls should get around 30% faster now.


burdiyan:
Please don't merge yet. This is not enough. We need to update all the CI manifests and Dockerfiles too, also the mise config. Also, we want Go 1.26.2, not 1.25.7.

burdiyan:
Handled by https://github.com/seed-hypermedia/seed/commit/762646c939d95f1a6ad40f0759ab691c3da7ecec.

PR #485, opened 2026-04-23
fix(onboarding): restore SKIP button on profile step in full-screen flow

horacioh:
Commit b60dfb2 ("Remove vault banner over whole app") gated the SKIP
button on the Profile step to modal mode only:

  <ProfileStep onSkip={modal ? handleSkip : undefined} ... />

That removed the ability to skip onboarding entirely from the main
full-screen flow — users landing on the Profile step had no way out
except completing account creation. The handleSkip handler, the SKIP
JSX inside ProfileStep, and the hasSkippedOnboarding persistence path
were all still intact; only the prop pass-through was gated.

Always pass handleSkip to ProfileStep. The skip path already:
  - sets hasSkippedOnboarding=true in the onboarding-v001 store
  - calls cleanupOnboardingFormData() to clear partial form state
  - invokes onComplete() so root.tsx hides the onboarding overlay

root.tsx:250 already respects hasSkippedOnboarding on subsequent
launches, and OnboardingDialog remains available for users who want
to set up an account later.

Add a regression test that renders Onboarding in non-modal mode,
advances to the Profile step, and asserts the SKIP button is present
and invokes the completion callback.

### DIFF
--- a/frontend/apps/desktop/src/__tests__/onboarding-flow.test.tsx
+++ b/frontend/apps/desktop/src/__tests__/onboarding-flow.test.tsx
@@ -190,7 +190,7 @@ vi.mock('@sentry/electron/preload', () => ({}))
 
 import {Onboarding} from '../components/onboarding'
 
-function renderComponent() {
+function renderComponent({modal = true}: {modal?: boolean} = {}) {
   const container = document.createElement('div')
   document.body.appendChild(container)
   const root = createRoot(container)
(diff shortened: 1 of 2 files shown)


ericvicenti:
Looks good to me!

Side note: the linked commit https://github.com/seed-hypermedia/seed/commit/b60dfb272d691aa1aa34b8ed63d880b0ea868436 seems to be unrelated the changes you are discussing here.

horacioh:
@ericvicenti can you test this manually and make sure it works?

PR #487, opened 2026-04-23
feat(editor): push to servers after publish via document machine actor

horacioh:
Restores push-on-publish for the unified editor, which was lost when publishing moved from the draft button into the document machine. A new pushDocument actor is spawned fire-and-forget from publishing.cleaningUp so the state transitions back to loaded immediately while the push runs in the background. The desktop consumer provides the actor, honoring the user's pushOnPublish preference and surfacing progress through a compact toast.


ericvicenti:
Could we please use PublishedToast to stay consistent with the other push behaviors? It gives more feedback about which servers have accepted the content.

frontend/apps/desktop/src/pages/desktop-resource.tsx
```diff
@@ -151,6 +159,15 @@ export default function DesktopResourcePage() {
...
   publishResourceRef.current = publishResource
 
+  // Push-on-publish plumbing: refs keep the fromPromise actor stable across
+  // renders while always reading the latest hook values when it fires.
+  const pushResource = usePushResource()
+  const pushResourceRef = useRef(pushResource)
+  pushResourceRef.current = pushResource
+  const pushOnPublish = usePushOnPublish()
+  const pushOnPublishRef = useRef(pushOnPublish)
+  pushOnPublishRef.current = pushOnPublish
```
ericvicenti: This code seems kind of strange. Sorry for the dumb question, but why are we doing this?

horacioh: The actor is created with useMemo(..., []) — empty deps, so it’s created once and keeps stable identity. Stable identity matters because:
- The actor is passed into documentMachine.provide({actors: {...}}) to build machine.
- machine is consumed by ResourcePage → useActor(machine) downstream.
- If machine identity changes, the running XState instance is recreated and the editing state resets (draft id, pending remote update, editing/publishing state — all gone mid-edit).

But a stable closure captures whatever values existed at the first render. usePushResource() and usePushOnPublish() return new references on every render (the function from usePushResource closes over useUniversalClient() and useGatewayUrl(), and usePushOnPublish() is a tRPC query whose object changes on data refresh). If we closed over them directly, the actor would keep calling stale versions — and wouldn’t see setting changes the user makes mid-session.

Refs bridge the two: a ref object has stable identity, but .current is live. The actor reads the latest values when it fires.

This is exactly the same pattern used three lines above for publishResourceRef, and writeDraftActor/publishDocumentActor use editorRef + publishResourceRef for the same reason.

ericvicenti: Ok makes sense. It feels strange to me still, but it seems like our tooling kind of forces this

PR #497, opened 2026-04-24
fix(desktop): edit profile writes Account Profile blob (fixes #494)

### LINKED ISSUE #494: Edit Profile Broken on Desktop

horacioh:
## Summary

- Fixes #494. Desktop `EditProfileDialog` now calls `grpcClient.documents.updateProfile` (writes the Account Profile blob) instead of `desktopUniversalClient.publishDocument` (which was appending a `setMetadata` change to the home document). Onboarding already did this correctly; the edit-profile flow had drifted.
- Adds a "Profile updated" success toast on both desktop and web.
- Adds Vitest coverage for the updated desktop dialog (4 cases: name-only edit preserves description, Blob icon triggers `fileUpload` → `ipfs://<cid>`, empty icon stays empty, spinner while loading — and asserts the success toast).
- Logs a follow-up in `.ai/todo.md` to migrate the web `updateProfile` helper off `publishDocument` (requires a client-signed Profile blob helper since browser-held `CryptoKeyPair` cannot call the server-managed `UpdateProfile` RPC).

The shared read path (`frontend/packages/shared/src/account-metadata.ts::accountMetadataFromAccount`, `api-account.ts::loadAccount`) already prefers `Account.profile.{name,icon,description}` over the deprecated `Account.metadata.*`, so once desktop writes to the Profile blob, `useAccount(uid)` reflects the update without any read-path change.

## Test plan

- [x] `./dev build-desktop`
- [x] `pnpm web:prod`
- [x] `pnpm test` (377 desktop unit + 4 new edit-profile-dialog cases pass)
- [x] `pnpm typecheck`
- [x] `pnpm format:write` (no diff)
- [ ] `pnpm audit` — fails, but only on pre-existing CVEs on `main` (electron <39.8.5, xmldom, ip, etc.); my commit has zero dep changes. Tracked in `.ai/todo.md` section 2.
- [ ] Manual: open desktop app, Edit Profile, change name + icon, save; confirm no new home-doc history entry, profile name/icon updates, success toast fires.

### DIFF
--- a/.ai/todo.md
+++ b/.ai/todo.md
@@ -67,3 +67,30 @@ The advisories don't reach runtime code — `xmldom` is pulled in via `@electron
 - [ ] Revisit `ip < 2.0.2` (SSRF, no patch) — reaches us via `lighthouse > puppeteer-core` and `react-devtools`. Both are devDeps only. Track whether upstream publishes a fix, or swap `react-devtools` for the browser-extension equivalent so we stop shipping it as a dep.
 
 **Estimate:** 30 min if option 1 works; half a day if option 2 is needed; no-go otherwise.
+
+## 3. Web Edit Profile — switch to Account Profile blob (own PR)
+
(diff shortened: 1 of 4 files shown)


ericvicenti:
I think it would be cleaner to create the blob on the frontend, like we do with comments. This way we will have more portability with web code.

If the TS SDK doesn't support profile writing, then it's a good idea to add support for it!

horacioh:
@ericvicenti we can cover this in the future, because I believe we should remove it completely. all profile editing should happen in the vault, not in this dialogs.

This PR is to at least fix that for now, but we eventually might remove this with a Vault-only flow.

PR #519, opened 2026-04-28
feat(rebase): implement draft auto-rebase on remote updates (Phase A)

horacioh:
## Summary

- **Three-way merge classifier** (`classifyRebase`, `computeTheirsTouches`, `applyRebasePlan`) in `document-changes.ts` — detects which blocks each side touched since the common base, identifies conflicts, and produces a merge plan
- **Document machine rebase region** — new parallel XState state (`rebase: idle | conflict`) with events `rebase.blockTouched`, `rebase.apply`, `rebase.detectConflict`, `rebase.dismiss`; persists `baseBlocks`, `mineTouchedIds`, `pendingRemoteDocument`, `pendingRebase` in context
- **ProseMirror touch-tracking plugin** in `document-editor.tsx` — records which BlockNote block IDs the user edits via `rebase.blockTouched` events; suppresses tracking during programmatic merges via `suppressChangeRef`
- **`useAutoRebase` hook** in `use-document-machine.ts` — debounced effect that fires when editing is idle + a pending remote update is stashed; runs classifier, applies silent auto-merge via surgical editor ops (`updateBlock`/`insertBlocks`/`removeBlocks`), restores cursor, shows a toast; sends `rebase.detectConflict` for conflicts (Phase B stub)
- **Priority polling** — `subscribeEntity` and `useResource` accept `priority: 'high'` which polls discovery every 3s while focused (vs 20s); active resource page now subscribes with `priority: 'high'`
- **Draft persistence** — `mineTouchedIds` and `baseBlocks` round-trip through the draft file so rebase state survives reload
- **`2-run-desktop` dev helper** — packages and launches a second Seed.app instance on alternate ports/appdata for two-instance testing
- **Tests** — new unit tests for `classifyRebase`/`computeTheirsTouches`/`applyRebasePlan` and document machine rebase transitions


ericvicenti:
A note on priority polling: I'm afraid it will not always work properly. For example say you have a sub to a document because there is a card embed. That is low pri. Then you click it and the new sub has high priority, but it uses the same subscription key so the app-sync will not re-create the sub and the polling will remain slow.

(feel free to hand this off to claude to see if this is a real issue)

ericvicenti (approved):
I haven't closely reviewed the rebase logic but it looks sensible.

Also please remove the .ai file

PR #540, opened 2026-05-03
fix(daemon): handle CORS preflight OPTIONS request for file upload endpoint

horacioh:
## Summary

- Fixes CORS preflight failure when pasting images or uploading files via the editor
- The `/ipfs/file-upload` endpoint sets CORS headers but rejects non-POST methods with 405, including the browser's required `OPTIONS` preflight
- Adds an early return for `OPTIONS` requests with `200 OK` before the POST-only validation
- Fixes both image paste and slash-menu image insertion, which both hit the same endpoint

## Testing

- `go vet ./backend/hmnet/...` — passes
- `go test ./backend/hmnet/...` — all tests pass


juligasa (approved):
I already commited changes this morning, mostly adding tests

ericvicenti:
I wonder why we need CORS for this endpoint? I thought we always upload content to the same host

PR #577, opened 2026-05-08
perf(desktop): cut entity discovery overhead via gating, dedup, blur pause, concurrency cap

horacioh:
## Summary

Investigation of a recent perf regression traced sustained daemon load to entity discovery: every `useResource` with `subscribed: true` opens a long-lived discovery loop on the daemon, and many call sites were either over-subscribing or fragmenting the renderer-side dedup key. With many joined sites + followed accounts + open documents, this multiplied into hundreds of independent `discoveryLoop()` invocations.

This PR addresses the issue along five axes, **one per commit**, so each can be bisected or reverted independently:

| # | Commit | What it does |
|---|---|---|
| 1 | `perf(frontend): drop subscribed:true from low-value views` | Remove active discovery from breadcrumbs, sidebar sites/profiles listings, notification rows, and a couple of nav chrome callers. These views need cached metadata, not active polling. |
| 2 | `perf(ui): gate embed subscriptions on viewport visibility` | New `useInViewport` hook (IntersectionObserver + 30 s grace) gates the three embed renderers so off-screen embeds in long documents don't each open a discovery loop. |
| 3 | `perf(frontend): normalize mismatched subscription options` | `doc-navigation` was forking a separate dedup key from the resource page's high-priority sub for the same id; `query-block` toggled `recursive` on display mode, forking the same target between two keys. Both fixed. |
| 4 | `perf(desktop): dedup entity subscriptions by id and fix stream cleanup` | The renderer ledger now keys per-entity (not per-`{id, recursive, priority, scope}`), merges callers' options up to the strongest, and owns one daemon sub + one discovery-state sub per entity. Fixes the latent bug where a second caller's discovery-state sub would be torn down by the first caller's cleanup. |
| 5 | `perf(desktop): pause entity subs while the window is blurred` | New `window-focus.ts` listens to renderer focus/blur. After a 30 s grace, all of this window's daemon subs and discovery-state subs are torn down; on focus they're re-issued from the preserved caller list. |
| 6 | `perf(desktop): cap concurrent discovery RPCs in the sync loop` | A small FIFO slot pool (`MAX_CONCURRENT_DISCOVERY = 4`) wraps `runDiscovery` so bursts of subs reaching their tick at once queue rather than pile in flight. |

Plus a follow-up `fix` commit for downlevel TS iteration and a `test` commit covering the dedup, merge, cleanup, and blur-pause invariants.

## Why each change is safe

- **Commit 1 / 3** — strict subset of previous behavior; the data is still queried via React Query, the entities the user is *actively* viewing are still subscribed by the resource page or by joined-site / followed-account derived subscriptions.
- **Commit 2** — embeds re-subscribe automatically when scrolled into view; grace period prevents thrash.
- **Commit 4** — daemon-side dedup already kept this from multiplying actual daemon work, so this is mostly fixing the renderer-side bookkeeping (and the cleanup bug). The behavior change: when callers disagree on `recursive`, the merged sub picks the strongest requested value (`recursive: true` if any caller wants it). If only non-recursive callers remain, the sub downgrades.
- **Commit 5** — preserves the caller list across pause, so re-focus restores subs without losing track of who needs them. The 30 s grace absorbs alt-tabs.
- **Commit 6** — backpressure only; never blocks indefinitely (slots free as RPCs finish). Cancellation re-checked before consuming a slot so unsubscribed loops don't still hit the wire.

## Test plan

- [x] `pnpm typecheck` (all 9 workspaces clean)
- [x] `pnpm test` — 422 tests pass, including 8 new tests under `frontend/apps/desktop/src/models/__tests__/entities.subscriptions.test.ts`:
  - Two callers on the same entity → one daemon sub + one discovery-state sub
  - Adding a recursive caller upgrades the merged sub
  - Removing the recursive caller downgrades it back
  - Discovery-state sub stays alive while any caller remains (the cleanup-bug regression test)
  - Rapid unmount/remount does not churn the daemon
  - Separate entity ids tracked independently
  - Blur grace period tears down then refocus re-issues
  - Refocus during grace cancels the pending pause without churn
- [x] `pnpm --filter @shm/ui test` (32 pass)
- [x] `pnpm --filter @shm/editor test` (110 pass)
- [x] `pnpm format:write` (no changes)
- [ ] **Manual smoke** before merging:
  - Open a heavy document with many embeds, scroll up/down, watch the sync footer panel — embeds should subscribe as they come into view and unsubscribe ~30 s after leaving.
  - Background and foreground a window — within ~30 s of blur, footer panel should empty; within ~1 s of focus, subs should reappear.
  - With several joined sites + followed accounts, observe daemon CPU / log output before vs. after this branch.

## Notes for review

- The `BlockEmbed*` exports now wrap an inner component in a single `<div ref={ref}>`. This adds one extra DOM element per embed but is necessary to host the IntersectionObserver target through all rendering branches. If layout shifts surface in design QA, we can switch to a `cloneElement` approach.
- `priority` and `scope` are kept on `EntitySubscription` but are no longer part of the dedup key, since the daemon `subscribe` wire format never accepted them.
- All commits are bisectable on their own — `pnpm typecheck && pnpm test` passes at every commit.

https://claude.ai/code/session_01RBiUf7sTa617hF4tybbswQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01RBiUf7sTa617hF4tybbswQ)_

### DIFF
--- a/frontend/apps/desktop/src/app-sync.ts
+++ b/frontend/apps/desktop/src/app-sync.ts
@@ -81,6 +81,34 @@ const DISCOVERY_POLL_INTERVAL_MS = 14_000
 const ACTIVITY_POLL_INTERVAL_MS = 1_000
 const DELETED_POLL_INTERVAL_MS = 60_000 // Slower polling for deleted/redirected resources
 
+// Cap concurrent in-flight discovery RPCs. The backend has its own worker pool
+// (config.MaxWorkers, default 6), but capping on this side keeps grpc-client
+// and tRPC overhead bounded when many subs reach a polling tick at once.
+const MAX_CONCURRENT_DISCOVERY = 4
+let discoveryInFlight = 0
+const discoveryQueue: Array<() => void> = []
+
+/** Runs `fn` immediately if a slot is free, otherwise queues it FIFO. */
+function withDiscoverySlot<T>(fn: () => Promise<T>): Promise<T> {
+  return new Promise<T>((resolve, reject) => {
+    const run = () => {
+      discoveryInFlight++
+      fn()
+        .then(resolve, reject)
+        .finally(() => {
+          discoveryInFlight--
+          const next = discoveryQueue.shift()
+          if (next) next()
+        })
+    }
+    if (discoveryInFlight < MAX_CONCURRENT_DISCOVERY) {
+      run()
+    } else {
+      discoveryQueue.push(run)
+    }
+  })
+}
+
 /** Returns 1 when app is focused, 10 when backgrounded. */
 function getPollingMultiplier(): number {
   return isAnyWindowFocused() ? 1 : 10
@@ -825,7 +853,12 @@ function createSubscription(sub: ResourceSubscription): SubscriptionState {
       return
     }
 
-    runDiscovery(sub)
+    withDiscoverySlot(async () => {
+      // Re-check cancellation before consuming a slot — the loop may have been
+      // unsubscribed while waiting in the queue.
+      if (cancelled) return null
+      return runDiscovery(sub)
+    })
       .then((result) => {
         if (cancelled) return
 
--- a/frontend/apps/desktop/src/components/doc-navigation.tsx
+++ b/frontend/apps/desktop/src/components/doc-navigation.tsx
@@ -13,7 +13,11 @@ export function DocNavigation({showCollapsed}: {showCollapsed: boolean}) {
   const route = useNavRoute()
   if (route.key !== 'document') throw new Error('DocNavigation only supports document route')
   const {id} = route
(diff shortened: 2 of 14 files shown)


ericvicenti:
Please, please. Lets only merge this if we are sure that it solves performance issues.

Because I know this will introduce many bugs and situations where content will not load.

The breadcrumbs for example- I recently introduced syncing to those, because the parent docs were not loading. And now I see that work being un-done and we will have that bug again.

Capping the discovery count- this will also cause some content to not load for the user. We don't know if this will help performance, but we do know it will cause regressions.

So, if this solves performance issues for users with 30 subscriptions, great, lets merge it and we can solve those regressions. Otherwise, please no

PR #580, opened 2026-05-10
Refine manual top navigation editing UX

GaboHBeaumont:
remove automatic top-nav behavior
make top nav manual-only
add empty-state CTA
simplify nav item editing into one surface
improve link picker UX and display path-only HM links
hide incomplete items from the top header


ericvicenti:
This looks good, but we need the tests to pass

horacioh:
## Code review fixes (commit `c6cc1534e`)

Applied 7 fixes raised in review on the nav editor work:

### 1. Popover open callback no longer mutates state unconditionally
`frontend/apps/desktop/src/components/edit-navigation-popover.tsx` — replaced the `usePopoverState(false, openCb)` form, which fired `editDocNav([createEmptyNavigationItem()])` on every open when nav was empty. **Why:** if a re-render landed between the popover opening and the state update, the empty check could pass twice and create two items. The fix moves the auto-add into a `useEffect` inside `EditNavigation`, guarded by a `didAutoAdd` ref so it can only fire once per mount.

### 2. Removed stale `autoFocusItemId` effect
Same file. The effect set `autoFocusItemId` once and never cleared, so the second blank item never received auto-focus. **Why:** the click handler in the Add-Item button already calls `setAutoFocusItemId(newItem.id)` directly, and the initial `useState(firstBlankItemId)` covers the mount case. The effect was redundant and actively wrong.

### 3. Drag drop zone now covers the whole card
Same file. Previously both `draggable` and `dropTargetForElements` were attached to the header row only, so dropping into an expanded item's form area missed. **Why:** added a separate `cardRef` on the outer card div for `dropTargetForElements` while keeping `dragHandleRef` (header row) as the drag source. Now you can drop onto any part of the card, including the expanded form.

### 4. Removed noisy `console.error` in drag setup
Same file. `console.error('Element ref not ready')` fired on the first paint before the ref attached — normal React behavior, not an error. **Why:** dev console pollution.

### 5. Stopped forwarding `directoryItems={[]}`
`frontend/apps/web/app/web-site-header.tsx` — `SiteHeader`'s `directoryItems` prop is already optional and the rendering block that consumes it is commented out. **Why:** passing an empty array was a semantic lie; if the prop ever comes back online a `[]` from this caller would silently render nothing.

### 6. Fixed vacuous test assertion
`frontend/apps/desktop/src/components/__tests__/edit-navigation-popover.test.tsx` — the "compact icon-only trigger" test was asserting `Add Navigation Item` appeared exactly once, identical to the empty-state test. **Why:** the intent is to verify the compact trigger does **not** contain that text. New assertion scopes to the trigger element via `.no-window-drag` and uses `not.toContain('Add Navigation Item')`. The inline button at the bottom of the popover still renders that text, so scoping to the trigger is necessary.

### 7. Cleaned up `computeHeaderData` signature
`frontend/packages/ui/src/resource-page-common.tsx` + `frontend/packages/ui/src/feed-page-common.tsx` — dropped `_siteHomeId` and `_directory` params, both unused (prefixed with `_` only to silence the warning). **Why:** all callers had to pass them anyway, and both call sites maintained a now-dead `homeDirectory = useDirectory(siteHomeId)` call solely to feed the ignored parameter. Removed the calls and the orphaned `useDirectory` imports.

---

**Verification:**
- `pnpm typecheck` — green across all 9 frontend workspaces
- `pnpm test` — desktop 421/421, shared 726/726, web 52/53 (1 pre-existing skip)
- `pnpm format:check` — clean

PR #594, opened 2026-05-12
improve some navigation issues

GaboHBeaumont:
### Summary
This PR improves the Navigation Settings popup so the Link field behaves like a real input instead of a static “Choose” control.

### What changed
Replaced the Link picker button with an editable combobox-style input
Users can now:
type to search internal documents
select search results directly from the dropdown
paste full external URLs and press Enter
The Link field now shows:
/path for internal hm links
the raw URL for external links
Selecting or entering a link now always updates the Label field
Preserved keyboard interactions:
↑ / ↓ to move through results
Enter to confirm
Esc to close/reset
Updated unit tests to cover:
editable Link input rendering
hm path display behavior
internal doc selection
pasted external URL handling
### Why
This makes the popup match user expectations: Label and Link both remain separate fields, but the Link field now feels like an actual input while still safely supporting internal document selection.

### How to test
Open the Navigation Settings popup and confirm Link is editable.
Type a document name in Link, select a result, and verify:
Link becomes /doc-path
Label updates to the selected doc title
Paste a full https://... URL into Link, press Enter, and verify Label updates too.
Open an existing internal nav item and confirm the Link field shows only the path, not the full hm://... value.
Check keyboard behavior: ArrowUp, ArrowDown, Enter, and Escape.

### DIFF
--- a/frontend/apps/desktop/src/components/__tests__/edit-navigation-popover.test.tsx
+++ b/frontend/apps/desktop/src/components/__tests__/edit-navigation-popover.test.tsx
@@ -1,8 +1,15 @@
-import React from 'react'
+import React, {useState} from 'react'
 import {createRoot, Root} from 'react-dom/client'
 import {act} from 'react-dom/test-utils'
-import {describe, expect, it, vi} from 'vitest'
 import type {HMNavigationItem} from '@seed-hypermedia/client/hm-types'
+import {afterEach, describe, expect, it, vi} from 'vitest'
+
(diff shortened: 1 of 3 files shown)


GaboHBeaumont:
@horacioh @ericvicenti can you review and merge pleaes? i want to include this in our release...

ericvicenti:
I think something went wrong with your rebase? I'm confused why this PR has changes to the notify service?

PR #602, opened 2026-05-13
desktop: restrict publish UI to edit context and grey disabled states

GaboHBeaumont:
### New publish-toolbar behavior
The desktop publish experience was updated so the Publish UI behaves more like an explicit edit action.

### Publish is only visible in Edit Context

- Hidden in Reading Context
- Hidden when a draft exists but the user is not actively editing
- Hidden on non-content/main non-edit views
- Publish button state is clearer

### Green when there are changes that can be published
Grey when there are no changes to publish
This applies both to:
the top-right Publish trigger
the main “Publish: Make it live now” button inside the popover
Publish popover stays as the main interaction

Still shows the publish dialog/details
Still allows access to status/info even when publish is unavailable
Publish status messaging improved

If the document has never been published, the popover shows:
“Not yet published”
This status is non-clickable
What was intentionally not added
No “highlight changes to publish” behavior
No change to the current outside-click / Esc editing behavior
No new publish confirmation flow; existing confirmation behavior remains

### DIFF
--- a/frontend/apps/desktop/src/components/__tests__/publish-popover-body.test.tsx
+++ b/frontend/apps/desktop/src/components/__tests__/publish-popover-body.test.tsx
@@ -202,6 +202,27 @@ describe('PublishPopoverBody', () => {
     }
   })
 
+  it('shows a non-clickable not-yet-published status for first publish docs', () => {
+    setSnapshot({
+      document: {version: '', metadata: {}},
+      draftId: 'abc',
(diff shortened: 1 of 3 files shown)


GaboHBeaumont:
Ready to review! I'd love have this one in the next release :)

ericvicenti (approved):
looks good!

GaboHBeaumont:
AWWWWWEESOME!!

PR #603, opened 2026-05-13
Add text placeholder only to the first block

GaboHBeaumont:
- Restricts the empty-editor placeholder so it appears only on the first top-level empty text block.
- Prevents the placeholder from appearing in the second or later empty blocks, even when those blocks are focused.
- Preserves the rule that the placeholder disappears once the first block has content and returns if that first block is cleared again.
- Keeps the behavior shared across document and comment editors, instead of introducing a document-only special case.
- Adds regression tests covering first-block-only placeholder behavior and the reappearance/absence cases.

### DIFF
--- a/frontend/packages/editor/src/blocknote/core/extensions/Placeholder/PlaceholderExtension.ts
+++ b/frontend/packages/editor/src/blocknote/core/extensions/Placeholder/PlaceholderExtension.ts
@@ -85,7 +85,8 @@ export const Placeholder = Extension.create<PlaceholderOptions>({
               const hasAnchor = anchor >= pos && anchor <= pos + node.nodeSize
               const isEmpty = !node.isLeaf && !node.childCount
               const isFirstTopLevelBlockContent = pos === firstTopLevelBlockContentPos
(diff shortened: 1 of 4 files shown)


frontend/packages/editor/src/blocknote/core/extensions/Placeholder/PlaceholderExtension.ts
```diff
@@ -54,29 +56,57 @@ export const Placeholder = Extension.create<PlaceholderOptions>({
             const active = this.editor.isEditable || !this.options.showOnlyWhenEditable
             const {anchor} = selection
             const decorations: Decoration[] = []
+            const topLevelBlockChildren = doc.firstChild
```
iskaktoltay: I would name this firstDocChild or something like that

iskaktoltay (approved):

frontend/packages/editor/src/blocknote/core/extensions/Placeholder/PlaceholderExtension.ts
GaboHBeaumont: Great.I will do.

PR #609, opened 2026-05-14
fix(shared): repair duplicate draft block ids before publish

### LINKED ISSUE #608: Failed to publish a document update
In production app I have this draft:

[yldYKQHQLK.json](https://github.com/user-attachments/files/27767298/yldYKQHQLK.json)

When I click publish, the console shows this error: `ConnectError: [unknown] block and left must not be the same`

Image

(This is Eric, btw 🤣)

ion-lion:
## Summary
- repairs duplicate editor block IDs before generating publish changes
- keeps the first occurrence stable and gives later duplicate draft blocks fresh IDs so duplicated/corrupted draft subtrees publish as new blocks
- adds a regression test that exercises the real `compareBlocksWithMap` publish-diff path and asserts it no longer emits `moveBlock.blockId === moveBlock.leftSibling`

Fixes #608.

## Before-fix reproduction
Reproduced on current `origin/main` (`b0e0cf664`) using the draft JSON attached to #608.

Command:
```bash
corepack pnpm exec tsx /home/e.guest/.openclaw/workspace/tmp/issue-608/repro.ts
```

Before the fix, `compareBlocksWithMap(createBlocksMap(baseBlocks, ''), draft.content, '')` produced one invalid move:

```json
{"blockId":"-X8B2_qS","parent":"X456eDA0","leftSibling":"-X8B2_qS"}
```

That matches the backend validation error from `backend/api/documents/v3alpha/docmodel/crdt_block_tree.go`: `block and left must not be the same`.

## Fix summary
The intended path is the shared publish-diff generation in `frontend/packages/shared/src/utils/document-changes.ts`.

This PR adds `ensureUniqueEditorBlockIds()` and runs it at the start of `compareBlocksWithMap()`:
- the first occurrence of each editor block ID is preserved;
- later duplicate occurrences get fresh `nanoid(10)` IDs;
- the existing diff logic then treats those duplicate copies as new blocks and emits valid `moveBlock`/`replaceBlock` changes.

## After-fix verification
Re-ran the same reproduction script against the patched branch:

```text
changes 295 touched 175
selfMoves 0
```

Also ran:
```bash
corepack pnpm --filter @shm/shared exec vitest --run src/utils/document-changes-rebase.test.ts
corepack pnpm --filter @shm/shared run typecheck
corepack pnpm --filter @shm/desktop run typecheck
git diff --check
```

## Self-review notes
- Patch is small and localized to the shared publish-diff layer.
- It does not relax or duplicate backend CRDT validation.
- It does not reimplement publishing; it repairs corrupt editor input before the existing diff builder runs.
- Normal non-corrupt drafts keep stable IDs because only repeated/missing IDs are regenerated.
- The regression test exercises the real diff helper, not a mocked publish result.

### DIFF
--- a/frontend/packages/shared/src/utils/document-changes-rebase.test.ts
+++ b/frontend/packages/shared/src/utils/document-changes-rebase.test.ts
@@ -1,6 +1,14 @@
+import type {EditorBlock} from '@seed-hypermedia/client/editor-types'
 import type {HMBlockNode} from '@seed-hypermedia/client/hm-types'
 import {describe, expect, it} from 'vitest'
-import {applyRebasePlan, classifyRebase, computeTheirsTouches} from './document-changes'
+import {
+  applyRebasePlan,
+  classifyRebase,
+  compareBlocksWithMap,
+  computeTheirsTouches,
+  createBlocksMap,
+  ensureUniqueEditorBlockIds,
+} from './document-changes'
 
 type BlockOverrides = {
   id: string
@@ -22,6 +30,10 @@ function para({id, text = '', revision, attributes}: BlockOverrides): HMBlockNod
 
 function node(over: BlockOverrides, children: HMBlockNode[] = []): HMBlockNode {
   return {block: para(over) as HMBlockNode['block'], children}
+}
+
+function editorPara(id: string, text = '', children: EditorBlock[] = []): EditorBlock {
+  return {id, type: 'paragraph', props: {}, content: [{type: 'text', text, styles: {}}], children}
 }
 
 describe('computeTheirsTouches', () => {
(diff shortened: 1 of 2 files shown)


horacioh:
Closing in favor of bf72a5a01 on `main`, which folds the structural improvements from this PR into the `deduplicateBlockIds` utility shipped in #610.

The follow-up commit:

- Moves the dedup call from the desktop `usePublishResource` call site **into `compareBlocksWithMap` itself**, so `@shm/web` and any other caller of the shared publish-diff helper get the fix automatically.
- Splits the recursion into a public `compareBlocksWithMap` (dedups once) and internal `compareUniqueBlocksWithMap` (recurses without re-dedup), matching the structure proposed here.
- Adds the collision-retry loop and empty/falsy-id handling from this PR's `nextUniqueBlockId` / `ensureUniqueEditorBlockIds`.
- Keeps the published-ID protection and 13 unit tests from #610, and adds end-to-end regression tests that drive real `compareBlocksWithMap` against duplicate sibling subtrees and assert no `moveBlock.blockId === leftSibling` is emitted — same shape as the assertion in this PR's regression test.

Thanks for the structural pointers — the diff-layer placement and the test-against-the-real-helper idea were the parts worth lifting wholesale.

ion-lion:
Thanks — that consolidation makes sense to me. Moving the repair fully into `compareBlocksWithMap` is the right layer, and folding it into the broader `deduplicateBlockIds` coverage from #610 is cleaner than keeping this PR alive separately. Glad the reproduction/test shape was useful.

PR #610, opened 2026-05-15
fix(documents): deduplicate block IDs before publish to prevent block==left CRDT errors

### LINKED ISSUE #608: Failed to publish a document update

horacioh:
## Summary

- Fixes `ConnectError: [unknown] block and left must not be the same` on publish (issue #608)
- Adds `deduplicateBlockIds` utility in `document-changes.ts` — walks editor blocks pre-publish and reassigns fresh 8-char `nanoid` IDs to any duplicates
- Published block IDs (from existing `blocksMap`) are "reserved": first encounter keeps original ID, subsequent duplicates get renamed
- Calls `deduplicateBlockIds` in `usePublishResource` mutation before CRDT change diffing
- Adds 13 unit tests covering: no-op case, sibling duplicates, nested duplicates, triple duplicates, deep nesting, published ID protection, and the original `block==left` scenario

---

Fixes #608

### DIFF
--- a/frontend/apps/desktop/src/models/documents.ts
+++ b/frontend/apps/desktop/src/models/documents.ts
@@ -36,6 +36,7 @@ import {queryKeys} from '@shm/shared/models/query-keys'
 import {
   compareBlocksWithMap,
   createBlocksMap,
+  deduplicateBlockIds,
   extractDeletes,
   getDocAttributeChanges,
 } from '@shm/shared/utils/document-changes'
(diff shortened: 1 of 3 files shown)


ericvicenti:
This seems good. Maybe we should also ensure unique block IDs when we are reading the draft from disk. So that way the user will see the exact state of the document that will be published (compared to the current user experience where I see two duplicated blocks)

horacioh:
> This seems good. Maybe we should also ensure unique block IDs when we are reading the draft from disk. So that way the user will see the exact state of the document that will be published (compared to the current user experience where I see two duplicated blocks)

after reviewing the code and the possible errors, doing this dedupe at publish time is the best solution to cover all edge cases and all cases in the publishing process. If we see this error again, we can come back to this review the new error

PR #626, opened 2026-05-18
Remove plus button from image caption

GaboHBeaumont:
Implemented the image-caption restriction so captions remain inline rich text only and no longer expose block-insertion UI or slash commands.

Notes
- No schema/content-model change was made.
- Existing plain-text caption behavior was preserved.
- Full local validation was blocked in this environment because frontend dependencies were not installed.


horacioh:
nice one @GaboHBeaumont

PR #649, opened 2026-05-20
Copy-pasting mentions (embed links) between documents does not work

GaboHBeaumont:
Yes — you’re right. If the copied thing is a mention / inline embed, paste should recreate a mention / inline embed, not show raw hm:// links or ask via the “Insert as” menu.

What was happening:

A copied mention is represented in HTML as an anchor like:
<a data-inline-embed="hm://..." href="...">Avatar options</a>

But a normal link is also an anchor:
<a href="...">Avatar options</a>

The generic link parser could claim the anchor before the inline-embed parser did, so the paste degraded into regular links / raw hm:// text.
I fixed that by:

giving a[data-inline-embed] parse rules higher priority
making the generic link parser explicitly ignore data-inline-embed anchors
keeping the prior clipboard serialization fixes so re-copying mentions still works

### DIFF
--- a/frontend/packages/editor/src/mentions-plugin.tsx
+++ b/frontend/packages/editor/src/mentions-plugin.tsx
@@ -10,6 +10,11 @@ import {Plugin} from '@tiptap/pm/state'
 import {NodeViewWrapper, ReactNodeViewRenderer} from '@tiptap/react'
 import './inline-embed.css'
 
+/** Fallback text used when serializing an inline embed to the clipboard. */
+export function inlineEmbedClipboardText(link: string): string {
+  return link || ''
+}
+
 /** Creates the TipTap Node for rendering inline-embed mentions in the document. */
 export function createInlineEmbedNode() {
   const InlineEmbedNode = Node.create({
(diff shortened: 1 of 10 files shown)


iskaktoltay (commented):
Why are there so many changed files? I have no idea what to look for. I believe it's not rebased cc. @horacioh

GaboHBeaumont:
@iskaktoltay 

I fix the commit, but now I realized that the AI messed Inline Embeds versus Mentions.

How do we differentiate between Inline Embeds and Mentions? We agreed that Mentions are strictly for mentioning People.

@horacioh @ericvicenti

horacioh:
Inline embeds and mentions in the code should be treated the same. we should not have two different types for the same thing

We can treat them differently in the UI, but in the permanent data should be the same

@GaboHBeaumont @iskaktoltay

horacioh:
@GaboHBeaumont You need to rebase your branch before you create the PR because if you don't rebase, you will include all the other code in your PR. so please rebase your branch first before pushing

horacioh (changes requested):
Please rebase your branch first before we accept this PR.

GaboHBeaumont:
Fixed now:

- rebased the branch onto current `main`
- force-pushed a clean history with a single commit
- narrowed the PR to only the inline-embed copy/paste fix (10 editor files)
- kept mentions / inline embeds as the same underlying type in code and data

Current head: `35a794dce` (`fix(editor): preserve inline embeds across copy-paste`)

Could you please take another look?

GaboHBeaumont:
Quick update:

- the PR is now rebased on current `main`
- the diff is narrowed to 10 editor-only files for the inline-embed copy/paste fix
- I re-requested review

One note on CI: the failing `integration-tests` job currently looks unrelated to this PR. It is failing during the web build on `frontend/apps/web/app/routes/hm.api.auth.test.ts` importing `action` from `hm.api.auth.tsx` (`MISSING_EXPORT` during the route build step).

Could you please re-review the current diff when you have a moment?

frontend/packages/editor/src/mentions-plugin.tsx
```diff
@@ -10,6 +10,11 @@ import {Plugin} from '@tiptap/pm/state'
 import {NodeViewWrapper, ReactNodeViewRenderer} from '@tiptap/react'
 import './inline-embed.css'
 
+/** Fallback text used when serializing an inline embed to the clipboard. */
+export function inlineEmbedClipboardText(link: string): string {
```
iskaktoltay, line 14: When copy pasting a mention in other platforms other than seed, the pasted text will be a raw hm link. Is this intended?

iskaktoltay (approved):
Approved with one comment

PR #694, opened 2026-05-27
fix(activity): use source document for comment mention notifications

### LINKED ISSUE #693: Mention notifications point to :profile or disappear after 2026.5.5
## User-facing impact

In release `2026.5.5`, mention notifications regressed compared with `2026.5.4`.

Observed behavior:

- **Web:** mention notifications do not appear on the notifications page.
- **Desktop:** mention notifications may appear, but the title is wrong, e.g.
  - `Alice mentioned you in :profile`
- This is confusing because the notification should point to the document/comment where the mention happened, not to the mentioned person’s profile resource.

Expected behavior:

- Mention notifications appear on both web and desktop.
- Notification copy should name the source document where the mention occurred, e.g.
  - `Alice mentioned you in Project Plan`
- Web site-scoped notification filtering should keep the notification under the source site/document, not filter it out because the payload target is the mentioned profile.

## Regression window

Working release: `2026.5.4`

Broken release: `2026.5.5`

Likely introducing commit:

- `ffd27345f perf(activity): speed up ListEvents reads ~250x in the unfiltered path`

That commit changed `backend/api/activity/v1alpha/activity.go`, especially the mention portion of `ActivityFeed.ListEvents`.

## What changed in the regression

The performance change introduced an unfiltered mention query path (`listMentionsCoreNoTargets`) and started surfacing `newMention` feed events more directly in the unfiltered activity feed.

For comment mention events, the activity feed payload currently has two different concepts:

- `newMention.target`: the entity being mentioned, e.g. the mentioned account/profile (`hm://<mentioned-account>/:profile`).
- `newMention.sourceDocument`: the document where the comment/mention happened.

The bug is that for `Comment` blobs, `ListEvents` was setting:

```go
sourceDoc = target
```

But in this query row, `target` is the mentioned entity from `resource_links.target`, not the comment’s target document.

So a comment that mentions a profile produces a feed event like:

```ts
newMention.target = "hm://<mentioned-account>/:profile"
newMention.sourceDocument = "hm://<mentioned-account>/:profile" // wrong
```

It should produce:

```ts
newMention.target = "hm://<mentioned-account>/:profile"
newMention.sourceDocument = "hm://<document-owner>/<document-path>" // correct
```

## Why this breaks web and desktop differently

The notify service consumes `ActivityFeed.ListEvents`.

For mention events, notification construction resolves the source document from:

```ts
const sourceHref = isCommentMention && mentionEvent.sourceDocument
  ? mentionEvent.sourceDocument
  : mentionEvent.source
```

Because `sourceDocument` was incorrectly set to the mentioned profile, the persisted notification payload was built with the mentioned profile as the notification target/source document.

Consequences:

- **Desktop:** the notification is present, but title generation uses the wrong target metadata/path, resulting in copy like `mentioned you in :profile`.
- **Web:** the notification page asks the notify service for site-scoped notifications. Site filtering uses `notification.target.uid`. Since the payload target is the mentioned account/profile rather than the site/document where the mention happened, the item can be filtered out and never appear on the web notifications page.

## Exact fix

In `backend/api/activity/v1alpha/activity.go`:

1. Extend the mentions SELECT to include the structural blob resource IRI:

```sql
source_resources.iri AS source_resource_iri
```

using:

```sql
LEFT JOIN resources source_resources ON source_resources.id = structural_blobs.resource
```

2. Read that column when constructing mention feed events.

3. For comment mention events, set:

```go
sourceDoc = sourceResourceIRI
```

instead of:

```go
sourceDoc = target
```

This preserves the correct separation:

- `Target` remains the mentioned entity/profile.
- `SourceDocument` becomes the document/resource containing the comment.

## Regression test

Add a backend regression test in `backend/api/activity/v1alpha/activity_test.go`:

```go
TestListEventsCommentMentionSourceDocumentUsesCommentTarget
```

The test inserts a synthetic comment blob whose resource is `hm://<author>/doc`, and a `comment/Embed` resource link targeting `hm://<mentioned>/:profile`.

It verifies that `ListEvents` returns:

```go
mention.SourceDocument == "hm://<author>/doc"
mention.Target == "hm://<mentioned>/:profile"
```

## Validation

Minimum validation:

```bash
go test ./backend/api/activity/v1alpha
```

Broader validation:

```bash
go test ./backend/...
```

Manual validation on staging:

1. Deploy patched backend + notify service.
2. Use a fresh mention event; existing persisted inbox rows may remain bad because notification persistence uses `INSERT OR IGNORE`.
3. From account A, mention account B in a comment on a document owned by account/site A.
4. Confirm raw activity feed has:
   - `newMention.target = hm://<B>/:profile` or `hm://<B>`
   - `newMention.sourceDocument = hm://<A>/<document-path>`
5. Confirm notify inbox payload target points to the source document/site, not account B’s profile.
6. Confirm:
   - web notifications page shows the mention
   - desktop title says `mentioned you in <document title>`, not `:profile`

## Note about existing broken rows

This fix corrects newly generated feed events and newly persisted inbox notifications. It may not repair already-persisted broken notification rows in the notify-service DB, because inbox insertion is currently `INSERT OR IGNORE` keyed by `(accountId, feedEventId)`.

If we need to repair existing user inboxes, follow-up work may be needed to backfill or rewrite bad persisted notification payloads whose mention target is `:profile` but whose feed event can still be resolved to the source document.

ericvicenti:
## Summary

Fixes #693.

In `ActivityFeed.ListEvents`, comment mention events were incorrectly setting `newMention.sourceDocument` to the mention target. For account/profile mentions, that meant the source document became `hm://<mentioned-account>/:profile`, causing notification titles like `mentioned you in :profile` and causing web site-scoped inbox filtering to hide the notification.

This PR keeps `newMention.target` as the mentioned entity, but sets `newMention.sourceDocument` from the comment blob's resource IRI.

## Changes

- Include the comment/source structural blob resource IRI in the mentions feed query.
- Use that source resource IRI for `SourceDocument` when the mention source blob is a `Comment`.
- Add a regression test covering a comment mention of `hm://<account>/:profile` inside `hm://<author>/doc`.

## Validation

```bash
go test ./backend/api/activity/v1alpha
```

### DIFF
--- a/backend/api/activity/v1alpha/activity.go
+++ b/backend/api/activity/v1alpha/activity.go
@@ -437,9 +437,10 @@ func (srv *Server) ListEvents(ctx context.Context, req *activity.ListEventsReque
 				blobID      = stmt.ColumnInt64(11)
 				observeTime = timestamppb.New(time.Unix(stmt.ColumnInt64(12), 0))
 				tsid        = blob.TSID(stmt.ColumnText(13))
-				//extraAttrs  = stmt.ColumnText(14)
-				isDeleted = stmt.ColumnText(16) == "1"
-				source    = stmt.ColumnText(18)
+				//extraAttrs        = stmt.ColumnText(14)
+				isDeleted         = stmt.ColumnText(16) == "1"
+				source            = stmt.ColumnText(18)
+				sourceResourceIRI = stmt.ColumnText(19)
 			)
 			genesisBlobIDs = append(genesisBlobIDs, strconv.FormatInt(stmt.ColumnInt64(17), 10))
 			if target == "" && blobType != "Comment" {
@@ -447,7 +448,7 @@ func (srv *Server) ListEvents(ctx context.Context, req *activity.ListEventsReque
 			}
 
 			if blobType == "Comment" {
-				sourceDoc = target
+				sourceDoc = sourceResourceIRI
 				source = "hm://" + author + "/" + tsid.String()
 				eventTime = timestamppb.New(tsid.Timestamp())
 
@@ -867,12 +868,14 @@ SELECT distinct
         WHEN structural_blobs.type != 'Change' THEN structural_blobs.genesis_blob
         ELSE coalesce(structural_blobs.genesis_blob, structural_blobs.id)
     END AS effective_genesis,
-	r2.iri AS source_iri
+	r2.iri AS source_iri,
+	source_resources.iri AS source_resource_iri
 FROM resource_links
 JOIN structural_blobs ON structural_blobs.id = resource_links.source
 JOIN blobs INDEXED BY blobs_metadata ON blobs.id = structural_blobs.id
 JOIN public_keys ON public_keys.id = structural_blobs.author
 JOIN resources ON resources.id = resource_links.target
+LEFT JOIN resources source_resources ON source_resources.id = structural_blobs.resource
 LEFT JOIN resources r2
   ON r2.genesis_blob = CASE
         WHEN structural_blobs.type != 'Change' THEN structural_blobs.genesis_blob
--- a/backend/api/activity/v1alpha/activity_test.go
+++ b/backend/api/activity/v1alpha/activity_test.go
@@ -3,6 +3,7 @@ package activity
 import (
 	context "context"
 	"fmt"
+	"seed/backend/blob"
 	"seed/backend/core"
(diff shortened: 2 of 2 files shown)


ericvicenti:
Looks like Julio solved this one https://github.com/seed-hypermedia/seed/commit/f8fe0f3cabf14c64aa758c9b4a6d72e09db1a943

PR #698, opened 2026-05-28
feat(frontend): keep documents visible during transient errors

horacioh:
## Summary
- Keep the document machine mounted after a document has loaded so transient resource errors, discovery flaps, or temporary not-found states do not blank the page.
- Store the last good document/version and transient resource status in machine context, with sync hooks and selectors for UI/debug consumers.
- Add a non-fatal header banner that explains refresh/discovery issues while continuing to render the last loaded document.
- Expose the React Query client in dev builds for web/desktop debugging and add document machine coverage for transient resource events.

### DIFF
--- a/frontend/apps/cli/cli-doc.md
+++ b/frontend/apps/cli/cli-doc.md
@@ -1,0 +1,3 @@
+## Hello new CLI doc
+
+first block here.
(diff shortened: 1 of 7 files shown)


ericvicenti:
Nice! Thank you for doing this- there was horrible experiences without this workaround.

I reallly wish we understood the underlying cause of the transient errors. But is probably related to the other issues we are working on (cc @juligasa @burdiyan )

PR #709, opened 2026-06-02
Fix weak bold styling in shared hm-prose typography

GaboHBeaumont:
Implemented the bold-visual fix across all shared .hm-prose surfaces by making semantic bold render at font-weight: 700, and added an editor E2E regression that checks rendered bold styling instead of only checking the ProseMirror mark.

Files changed: frontend/packages/ui/src/hm-prose.css, frontend/packages/editor/e2e/test-app/TestEditor.tsx, frontend/packages/editor/e2e/tests/selection-toolbar.e2e.ts
This should affect editor, viewer, embeds, and other surfaces that share .hm-prose
Validation was attempted but blocked in this worktree because the local JS dependencies/toolchain are not fully installed (tsc/prettier unavailable)

### DIFF
--- a/frontend/packages/editor/e2e/test-app/TestEditor.tsx
+++ b/frontend/packages/editor/e2e/test-app/TestEditor.tsx
@@ -1,5 +1,6 @@
 import '@/blocknote/core/style.css'
 import '@/editor.css'
+import '@shm/ui/hm-prose.css'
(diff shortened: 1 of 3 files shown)


iskaktoltay (approved):
I don't think single line changes like this require a whole PR

PR #738, opened 2026-06-07
handle copy-pasting images when there is CORS

GaboHBeaumont:
## Summary

Fixes rich HTML paste from webpages where copied content includes remote images that are blocked by renderer-side CORS fetches.

Instead of trying to fetch remote `http/https` image URLs directly in the paste plugin, the editor now lets those images go through the normal HTML paste flow. That allows the existing desktop `importWebFile` path to import the images through the app/backend path, avoiding browser CORS restrictions.

## Changes

- Only convert pasted HTML image sources to `File`s when they are local browser sources:
  - `data:`
  - `blob:`
- Let remote image URLs fall through to the normal HTML paste/import flow:
  - `http://...`
  - `https://...`
- Wire desktop `importWebFile` into additional `CommentEditor` surfaces:
  - document comments
  - inline comment editing
  - agent message composer
  - agent prompt editor
- Reuse the shared editor import/file-attachment types in `CommentEditor`.
- Add unit coverage for local-vs-remote pasted image source handling.

## Why

Copying from pages like https://xanadu.com/zigzag/ can put remote image references into the clipboard HTML. The previous paste path tried to fetch those image URLs from the renderer, which fails when the source site does not send permissive CORS headers.

This caused paste to fail with errors like:

```txt
Access to fetch at 'https://xanadu.com/zigzag/zzlogo.jpg' from origin 'http://localhost:17654' has been blocked by CORS policy
Error processing pasted HTML image: TypeError: Failed to fetch

### DIFF
--- a/frontend/apps/web/app/routes/hm.api.web-file.tsx
+++ b/frontend/apps/web/app/routes/hm.api.web-file.tsx
@@ -1,0 +1,105 @@
+import type {LoaderFunctionArgs} from '@remix-run/node'
+import {MAX_FILE_SIZE_B, MAX_FILE_SIZE_MB} from '@shm/shared/constants'
+import {lookup} from 'dns/promises'
+import net from 'net'
+
+const MAX_REDIRECTS = 5
+
+export async function loader({request}: LoaderFunctionArgs) {
+  const urlParam = new URL(request.url).searchParams.get('url')
+  if (!urlParam) return new Response('Missing url', {status: 400})
+
+  try {
+    const response = await fetchRemoteFile(urlParam)
+    const contentLength = Number(response.headers.get('content-length') || '0')
+    if (contentLength > MAX_FILE_SIZE_B) {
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

**Comment by ericvicenti** [comment]
Ok so this fix is only meant to help the bug on desktop?

**Comment by GaboHBeaumont** [comment]
You are right! It only fixed Desktop.. I pushed a fix for web too.

File: frontend/apps/web/app/routes/hm.api.web-file.tsx [hunk 1]
```diff
@@ -0,0 +1,105 @@
...
+        all: true,
+        verbatim: true,
+      })
+
+  if (addresses.some(({address}) => isPrivateAddress(address))) {
+    throw new Error('URL host is not allowed')
+  }
+}
+
+function isPrivateAddress(address: string) {
```
ericvicenti, line 80: this is kind of a red flag, indicates a larger security issue with this approach. Probably we should not have this `web-file` interface at all. The client should request the file and upload it normally.

Is this intended as a sort of CORS bypass?

PR #740, opened 2026-06-08
fix(frontend): allow draft editing on latest version routes

horacioh:
## Summary
- Allow writers to edit drafts when viewing routes pinned to the current latest version.
- Preserve immutable behavior for true historical version snapshots.
- Refresh draft, draft list, and activity feed caches after publishing so updated versions appear promptly.


horacioh:
@GaboHBeaumont this should fix the "i can't edit documents I'm allowed to" issue.

PR #748, opened 2026-06-09
feat(syncing): maintained RBSR fingerprint index for fast blob propagation

juligasa:
## What

Replaces the per-round RBSR store rebuild with a **maintained, persistent fingerprint index** so an interested, already-connected peer gets a new blob in ~1s without melting CPU.

Today every reconcile round (and every peer) rebuilds the entire sorted blob set from SQLite via `collectBlobs`. This PR persists each scope's resolved set, serves range fingerprints from an O(log N) monoid tree, keeps the set current incrementally as blobs are indexed, and drops the hot discovery cooldown to 1s.

## How

- **Monoid tree** (`rbsr.NewTreeStore`): an augmented treap with deterministic hash-derived balancing. `RangeFingerprint` is O(log N) and **byte-identical** to the existing linear fold (the fold never sets the accumulator length, so the tree doesn't either). Supports post-seal insert. Wired via an optional `RangeFingerprinter` seam in `Session.Fingerprint` (falls back to the fold otherwise).
- **Persistence** (`rbsr_scope` / `rbsr_item` + migration): a scope = a canonical discovery key + protocol version. We persist the *resolved blob set* (not literal tree nodes — a treap with rotations doesn't map to stable rows) and rebuild the in-memory tree from those rows. Materialization runs `collectBlobs` once; reindex drops the tables (lazy re-materialize).
- **`loadStore`** serves from the persistent index, with a reindex-gated fallback to the legacy rebuild.
- **Oracle** (`affectedScopes`): the incremental membership decision — which materialized scopes a freshly indexed blob joins. **Seed-and-forward**: only the *seeding* decision is new logic; the closure reuses `collectBlobs`' own forward walk, so it can't silently diverge. Covers the dominant document-edit path (a new Ref + the Changes it heads via the media closure).
- **Index hook** (`Index.SetIndexedHook`): patches materialized scopes inside the indexing transaction, so the maintenance commits atomically with the blob.
- **Shadow-verify**: a periodic sweep recomputes each scope's set the authoritative way and re-materializes any that drifted — the safety net for the edges the oracle defers (inbound Contact-by-subject, capability delegation, late link targets).
- **`/0.9.3` canonicalization primitives**: `canonicalCodecFor` (dag-pb → raw), scope keyed by protocol version, build-from-rows applies it. Unit tests prove a dag-pb blob advertises identically to a raw-storing peer.
- **Hot cooldown** hardcoded to 1s.

## Deferred (NOT in this PR): live `/0.9.3` protocol negotiation

The canonicalization *primitives* are in and tested, but the live wire negotiation is **deliberately left out** and must be a separate, e2e-tested change. I implemented it, found a concrete breakage, and reverted it:

- `connect.go` gates connections with `CheckHyperMediaProtocolVersion(..., n.protocol.Version)`, requiring the peer's `FindHypermediaProtocol` result to **exactly equal** `0.9.2`. Peerstore `GetProtocols` returns a map (nondeterministic order), so advertising both `0.9.2`+`0.9.3` would intermittently surface `0.9.3` and **reject the peer**.
- Activating it needs: rewriting `FindHypermediaProtocol`/`CheckHyperMediaProtocolVersion` for highest-mutual negotiation, a second listener/server tagging the canonical context, and client-side dial selection + a canonicalized client store. That touches connection establishment and must be validated with the daemon e2e harness.

Until then the index serves the legacy-codec set (identical fingerprints to the old path), so this PR is backward-compatible.


backend/storage/schema.sql
```diff
@@ -376,4 +376,40 @@ CREATE INDEX fts_index_by_genesis_blob ON fts_index (genesis_blob);
...
+    -- 1 once materialized from collectBlobs; 0 means it must be re-materialized
+    -- before serving (first use, post-reindex, or repair).
+    materialized INTEGER NOT NULL DEFAULT 0,
+    -- Unix seconds of last access, used to evict idle scopes from the working set.
+    last_access INTEGER NOT NULL DEFAULT 0,
+    UNIQUE (iri, recursive, depth_one, blob_types, protocol_version)
+);
+
+-- The resolved blob set for each maintained scope.
+CREATE TABLE rbsr_item (
```
burdiyan, line 408: I guess you didn't end up using the fancy tree approach, and we simply store blob <-> scope relationship in a table?

backend/storage/schema.sql
```diff
@@ -376,4 +376,40 @@ CREATE INDEX fts_index_by_genesis_blob ON fts_index (genesis_blob);
...
+-- rbsr_item holds the resolved blob set for the scope, maintained incrementally
+-- by the syncing oracle so reconciliation no longer rebuilds the set per round.
+CREATE TABLE rbsr_scope (
+    id INTEGER PRIMARY KEY,
+    -- The discovery key's resource IRI.
+    iri TEXT NOT NULL,
+    -- Whether the scope covers the entire subtree below the IRI.
+    recursive INTEGER NOT NULL DEFAULT 0,
+    -- Whether the scope covers only the direct children of the IRI.
+    depth_one INTEGER NOT NULL DEFAULT 0,
```
burdiyan: I never understood this depth one thing. Why do we limit to one? Also, doesn't it conflict with recursive? Maybe we should use it as an enum instead of two booleans?

backend/storage/schema.sql
```diff
@@ -376,4 +376,40 @@ CREATE INDEX fts_index_by_genesis_blob ON fts_index (genesis_blob);
...
+    -- Canonical (sorted, comma-joined) blob-type allowlist. Empty means no filter.
+    blob_types TEXT NOT NULL DEFAULT '',
+    -- Wire protocol version whose codec canonicalization this scope's set reflects.
+    protocol_version TEXT NOT NULL,
```
burdiyan: Do you think it's a good idea to tie scopes to protocol versions? I could imagine wanting to bump the protocol version without needing to update the scopes.

Also this seems to have the ability to make migrations more complicated, because now we need to make sure to regenerate the scopes when we bump the protocol version.

Is there any practical use case we know of for this?

backend/hmnet/syncing/rbsr/monoidtree.go
```diff
@@ -0,0 +1,258 @@
...
+
+// NewTreeStore creates a [Store] backed by a monoid tree. It additionally
+// implements [RangeFingerprinter].
+func NewTreeStore() Store {
+	return &treeStore{
+		seen: make(map[string]struct{}),
+	}
+}
+
+func sizeOf(n *treeNode) int {
```
burdiyan: Why not making this a method on treeNode?

backend/hmnet/syncing/scheduler.go
```diff
@@ -19,7 +19,7 @@ import (
 
 const (
 	defaultHotTTL      = 40 * time.Second
-	defaultHotCooldown = 20 * time.Second
+	defaultHotCooldown = 1 * time.Second
```
burdiyan: Didn't we say 3 seconds? Actually yesterday on the call @ericvicenti mentioned to start with 10 seconds at least.

ericvicenti: 3 seconds is the goal for load time of any remote content

10 seconds is the goal immediacy for content updates while it is visible on your screen

burdiyan:
> the fold never sets the accumulator length, so the tree doesn't either

@juligasa this is actually an oversight I've been meaning to improve for a long time, and bumping the protocol version. Maybe we should take advantage of this change and do it?

It should also help us with measurements a little bit, because we'll avoid new peers talking to old peers as it could show quite different performance characteristics.

backend/storage/schema.sql
```diff
@@ -376,4 +376,40 @@ CREATE INDEX fts_index_by_genesis_blob ON fts_index (genesis_blob);
...
\ No newline at end of file
+);
+
+-- Maintained RBSR fingerprint index.
+-- Each rbsr_scope is one reconciliation scope: a canonical discovery key
+-- (IRI + recursion flags + blob-type allowlist) plus the wire protocol version,
+-- since different protocol versions advertise different codec-canonical sets.
+-- rbsr_item holds the resolved blob set for the scope, maintained incrementally
+-- by the syncing oracle so reconciliation no longer rebuilds the set per round.
+CREATE TABLE rbsr_scope (
```
burdiyan, line 391: Does it mean that every document will produce multiple scopes by default? E.g. the regular one, the recursive one, the depth-one one, and maybe some else, e.g. combining different blob types?

And for each of these scopes there's going to be tons of rows in `rbsr_items` table?

Maybe we should quit having filters per scope, and for things that we need filters for we simply define different scopes. This should simplify the whole thing by flattening it into one dimensions, instead of having multiple.

backend/hmnet/syncing/syncing.go
```diff
@@ -506,9 +506,39 @@ func (s *Service) Run(ctx context.Context) error {
...
 }
 
+// shadowVerifyInterval is how often the maintained RBSR index is checked against
+// a fresh collectBlobs materialization. Drift (a scope whose incrementally
+// maintained set diverged — e.g. via an edge the oracle defers) forces that
+// scope to re-materialize, which is the safety net for the deferred edges.
+const shadowVerifyInterval = 5 * time.Minute
+
+func (s *Service) runShadowVerify(ctx context.Context) {
+	ticker := time.NewTicker(shadowVerifyInterval)
```
burdiyan: I think this should be a timer, not a ticker. Actually I can't remember having a legitimate use case for a Ticker in my entire career.

Ticker keeps ticking after it fires, no matter how long your work is taking, but most periodic processes want "do work, sleep for some time, do work again", and for these scenarios a timer is always better. Just start a timer with a 0 so it fires immediately, and the reset it to the desired interval after each work cycle.

backend/hmnet/syncing/shadow_verify.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil, err
+	}
+	return rows, nil
+}
+
+// ShadowVerifySweep runs shadow-verify across every materialized scope, each in
+// its own write transaction so the sweep never holds the writer lock for the
+// whole pass. Returns how many scopes were checked and how many drifted (and
+// were marked for re-materialization). Driven by a periodic background loop.
+func ShadowVerifySweep(ctx context.Context, db *sqlitex.Pool) (checked, drifted int, err error) {
```
burdiyan: I think this function is not used anywhere outside its package, so it should probably be private?

backend/hmnet/syncing/canonical.go
```diff
@@ -0,0 +1,62 @@
...
+	if v, ok := ctx.Value(protocolVersionCtxKey{}).(string); ok && v != "" {
+		return v
+	}
+	return ProtocolVersionLegacy
+}
+
+// Wire protocol versions relevant to RBSR set advertisement.
+//
+// Under the legacy version a blob is advertised under whatever codec it was
+// stored with, so the same content held as raw on one peer and dag-pb on
```
burdiyan: Are you sure all of this is not an AI hallucination? Why would a dag-pb blob be stored as raw anywhere? Or vice versa?

Raw is raw, and dab-pb is dag-pb and there should be no way to confuse this anywhere. There should be no need to do any magic like this. Let's just bump the protocol, repair whatever needs to be repaired, and be done with it.

All this "backward-compatibility" code does indeed look like something Codex on high thinking level would do tbh 😅.

There should be no way to misinterpret a blob whether it's raw or dag-pb.

backend/hmnet/syncing/syncing.go
juligasa: yeah you right, a ticker doesnt really fit here. changed it to a timer that fires at 0 and then i reset it after every sweep, so now its work -> sleep -> work like you said. also the first check runs at startup now which is a nice side effect.

backend/hmnet/syncing/shadow_verify.go
juligasa: true, its only called inside the package. made it private now -> shadowVerifySweep.

backend/hmnet/syncing/canonical.go
juligasa: ok so before removing it i wanted to be sure, so i checked on a real node. let it sync ~9gb from the network and looked at the codec-mismatch counter in /debug/network -> totally empty, 0 mismatches even after 130k+ dropped blocks. so yeah the raw vs dag-pb confusion basically never happens, you were right.

and the other thing is this code was not even doing anything yet -- the protocol negotiation that would turn it on is deferred, so codecForProtocol was always the identity. so i just removed the whole canonical.go. if we ever really need it, it should land together with the 0.9.3 bump + reindex, not sit here as dead code. good catch.

burdiyan:
@juligasa 

Looks like the indexing hook misses blobs indexed by unstash cascades: https://github.com/seed-hypermedia/seed/blob/10d9ec37cb474a301a3cb41f7134b13323b29dab/backend/blob/index.go#L1947 calls `indexBlob` for previously stashed blobs during `reindexStashedBlobs`, but those blob IDs are never passed to `idx.runIndexedHook`. The only hook calls are in `Put` / `PutMany` for the top-level fresh IDs it seems: https://github.com/seed-hypermedia/seed/blob/10d9ec37cb474a301a3cb41f7134b13323b29dab/backend/blob/index_blockstore.go#L44 and https://github.com/seed-hypermedia/seed/blob/10d9ec37cb474a301a3cb41f7134b13323b29dab/backend/blob/index_blockstore.go#L104. So when a missing dependency/capability arrives and unstashes older refs/changes, the legacy tables get updated but `rbsr_item` does not. Reconcile can then serve a stale-short maintained index.

burdiyan:
@juligasa 

1. Don't we want to bump p2p protocol version? https://github.com/seed-hypermedia/seed/blob/1b52e543157f455da3e4c79674b9e988b74d11bf/backend/hmnet/hmnet.go#L59

2. I'm really concerned with this every 5 minute full sweep: https://github.com/seed-hypermedia/seed/blob/1b52e543157f455da3e4c79674b9e988b74d11bf/backend/hmnet/syncing/syncing.go#L529-L556

Isn't it super wasteful to re-evaluate all the scopes every 5 minutes? I mean, it really doesn't make sense doing it if there's no new blobs for example. It seems like we are not really confident in that we are able to accurately detect which blobs affect which scopes, so we have to do this safety net thing every 5 minutes to make sure we don't mess it up.

I don't mind if we run it for a while, but having it long-term doesn't seem right. And it doesn't seem right to run it when we can know for sure there's nothing that could have changed the world (e.g. no new blobs since the last sweep).

burdiyan:
I just realized something that made me think that we definitely should bump the protocol version, and change the signature of the ReconcileBlobs RPC in `syncing.proto`. We must adapt it to this new notion of scopes, and we must avoid combining multiple scopes into a single range, like we currently do by allowing multiple `filters`. This will break all the performance improvements from having well-defined scopes, I think.

burdiyan:
Here's a review finding from my agent (GPT 5.6 Sol). It looks valid, especially if we do something about [my previously shared concerns about the need for the background shadow sweep I shared above](https://github.com/seed-hypermedia/seed/pull/748#issuecomment-4902780478).

> Preserve missing Raw targets across materialization
>
> Filtering out `size < 0` placeholders here loses late Raw media targets permanently from this scope: Raw blocks bypass `indexBlob` and therefore never reach `MaintainRBSRIndex` when their data later arrives. The authoritative set will contain the Raw blob after download, but the maintained set stays short until shadow verification runs and another reconciliation re-materializes it. Either retain placeholder membership (filter only while serving) or invoke maintenance for newly downloaded non-indexable blocks.

PR #774, opened 2026-06-15
fix(vault): proxy IPFS blobs through vault server

GaboHBeaumont:
Fixes dev vault profile publishing by routing IPFS blob reads/writes through the vault server.

Adds /vault/ipfs/:cid GET/POST proxy routes to forward blob requests to the configured daemon backend.
Updates the vault frontend to use same-origin /vault/ipfs/{cid} instead of depending on public /ipfs/* routing.
Sets Docker defaults so the vault container talks to seed-daemon:56001 internally.
Validation:

bun run tsgo --noEmit
bun test (196 pass)

### DIFF
--- a/vault/Dockerfile
+++ b/vault/Dockerfile
@@ -17,5 +17,7 @@ WORKDIR /app
 COPY --from=build /build/vault/dist/ .
 VOLUME /data
 ENV SEED_VAULT_DB_PATH=/data/vault.sqlite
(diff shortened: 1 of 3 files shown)


burdiyan:
The dev server was down, hence the vault couldn't write the profile data to the backend. It's up now, and I just tested it, and it seems to work fine, so I'm closing this.

PR #839, opened 2026-07-01
Identify daemon signing keys by public key

ericvicenti:
## Summary

This PR removes daemon-side "named keys" and identifies signing keys by their public key instead.

The previous API exposed a key "name" even though the name was usually just the key's public key/account principal. That made the model look flexible while most call sites assumed the name and public key were interchangeable. In the uncommon cases where a key was registered with a custom name, those assumptions broke down and the daemon could fail with `named key not found` because some paths resolved by name while others reasoned about the public key.

## What changed

- Drops key names from the daemon key API and generated Go/TS clients.
- Uses `public_key` / `signing_key` as the key identifier for signing, export, and deletion.
- Removes the key rename RPC and keystore rename plumbing.
- Keeps existing vaults compatible by resolving stored accounts by either legacy name or principal where needed.
- Updates desktop/web/vault callers to pass public keys directly instead of maintaining account-id-to-key-name mappings.
- Leaves separate user-named identity systems (CLI keyring and agents-server identities) unchanged.

## Notes

- Stacked on #827 (`feat/account-settings`); this PR is intended to review the key identity cleanup separately.

## Testing

- Not run by me while publishing this PR.


burdiyan:
As I mentioned in Discord I think remove named keys is a bit risky, and I'd rather fix the bug where we assume the name is a public key instead of what the daemon tells the name is. Key names have "leaked" into the Vault data model, so it's a possibility that there's a Vault out there with a custom name that is not a public key. In this case I'm not sure what's going to happen to be honest when they attempt to sync up with the desktop app.

Unless there's other bugs I'm not aware of, it should be much easier to properly pass the name around in the client code than trying to remove it. I know you hate it, I'm not a fan of it anymore either, but it seems dangerous, and I'm too busy with other stuff to give this the attention it deserves.

burdiyan:
I pushed this https://github.com/seed-hypermedia/seed/commit/394d3e108c501d4d2c127069b3b80d47bad1d672.

Now we key name is not found we fall back to trying to lookup by public key. This should fix the "named key not found" errors without having to make huge changes right now.

ericvicenti:
Ok thanks, that seems reasonable for now

PR #840, opened 2026-07-01
Add table block to the editor

iskaktoltay:
Adds a full Table block to the Seed editor. 

### Data model
- A Table block holds TableColumn blocks and TableRow blocks as children, and each cell is an ordinary Paragraph block carrying a columnId that references its column.
- Row order is the order of TableRow blocks and column order is the order of TableColumn blocks.
- Cell position is its row and its columnId reference, not physical sibling order. 
- Headers use an isHeader attribute on TableRow/TableColumn, derived from the PM tableHeader vs tableCell node type. A position-0 invariant is enforced by the converters. Only row 0 / column 0 may be a header.
- New Zod schemas: HMBlockTableSchema, HMBlockTableRowSchema, HMBlockTableColumnSchema.

### Editor
- Vendored tiptap table extension and extended it with a columnId attribute on cell/header PM nodes and stable IDs on cells and rows for comment ranges.
- Class-based table NodeView.
- Row/column control hover strips and action menu (insert/delete row/column, toggle header row/column). Header auto-promote: inserting above/left of a header inserts the new header and demotes the old one in a single transaction, preserving the perpendicular header's corner cell.
- Keyboard: Enter moves down a cell, Shift-Enter inserts a hard break, Backspace/Delete is absorbed at cell boundaries.
- Custom HTML table paste handler.

### Conversions
Full round-trip across all three representations (HMBlockNode, editor block, ProseMirror), including header derivation, columnId propagation, orphan-cell drop, missing-cell synthesis, and the position-0 invariant.

### Comments & quotes
A comment/quote anchors to a cell's ID and character range. Cell IDs survive insert/delete/toggle/round-trip, so anchors don't orphan. The block-highlight plugin now recognizes tableCell/tableHeader nodes, so cell-anchored highlights render.

### Tests
- 28 table command tests: insert/delete row+column, header auto-promote, toggle header, and id/columnId preservation across ops.
- direct PM to editor-block converter tests: header derivation, position-0, colwidth, missing-cell synthesis, orphan drop, round-trip.
- Editor block to HM block tests: round-trips, position-0 normalization, orphan tableRow/tableColumn filtering.


horacioh (approved):
Fixed both review findings: table cell conversion now preserves paragraph-level props such as revision when rebuilding the inner PM paragraph, while keeping columnId on the table cell/header attrs where it belongs. I also added a regression test to cover this behavior so published table-cell text remains referenceable for comments and fragment links. Separately, I updated the formatting toolbar test mock to export getNearestBlockOrCellPos, matching the helper now used by the toolbar. Targeted tests, typecheck, and formatting checks pass.

PR #851, opened 2026-07-08
SSR refactor: render real editor markup server-side (pixel-parity, no manual sync)

ericvicenti:
## Why

The web app's pre-hydration document HTML came from a hand-written string renderer (`ssr-render.ts`) that **imitated** the editor's DOM. Every block type, class name, and layout quirk had to be duplicated by hand and kept in sync with the editor forever — and it had drifted: measured against 12 real pages (seedteamtalks site), the server paint differed from the hydrated paint by **1–4.6% of pixels on every single page**. Users saw it as content flashing and layout jumping on load.

## What

The SSR HTML is now generated from **the same sources of truth the mounted editor uses**, so drift is structurally impossible:

1. **Same conversion** — `hmBlocksToEditorContent` → `blockToNode` → the identical ProseMirror doc recipe as `BlockNoteEditor.onBeforeCreate`.
2. **Same scaffolding** — `DOMSerializer.fromSchema(getSchema(extensions))` executes the schema's own `renderHTML` headless in Node (happy-dom), emitting the editor's real hashed CSS-module classes. Those stylesheets now load eagerly (`@shm/editor/editor-styles`) instead of only inside the lazy editor chunk.
3. **Same components** — React node-view blocks (query, embed, image, video, code chrome, unknown…) render via `react-dom/server` with the loader's prefetched React Query cache and the same provider stack the page uses. Content-view embeds recurse through the same pipeline.

The editor still mounts exactly as before and swaps the placeholder in one commit — with identical markup the swap is invisible, and the editable↔read-only lifecycle is untouched.

## Measured result

`apps/web/scripts/ssr-parity-check.mjs` (committed, standing harness) screenshots each page with JavaScript disabled vs fully hydrated and pixel-diffs them:

| Page set (12 real pages) | Before | After |
|---|---|---|
| Worst page (`/design`, 12 embeds) | 4.57% | **0.001%** |
| Query-block pages (`/`, `/notes`, `/issues`) | 1.0–2.5% | **0.007–0.021%** |
| Code/image-heavy docs | 0.9–2.6% | **0.000–0.001%** |

The residual 0.0x% is live-data churn (relative timestamps ticking between the two captures). Spot-checked at 900/768px widths too.

## Real bugs fixed along the way

- **Query-block prefetch never worked**: the loader's cache key didn't match the component's `useQuery` key, so prefetched data was never found — the original "query block flashes in" bug. Key derivation is now single-sourced (`query-block-input.ts`).
- **Content blanked after editor mount**: `LazyViewportMount` waited for an IntersectionObserver round-trip before showing content SSR had already painted (~100–300ms collapse, DOM traced 845→235→877 elements). A pre-paint layout-effect check eliminates it; below-viewport blocks still lazy-mount.
- **One-frame blank at the editor swap**: the editor's React node views flush a frame after the ProseMirror DOM mounts, so the placeholder→editor commit painted empty block containers for a frame (caught via CDP screencast frame-diffing). `DocumentContentHandoff` keeps the placeholder stacked over the mounting editor for two animation frames — the screencast now shows zero visual changes at the swap.
- **Selection ring on read-only pages**: documents starting with a selectable block showed ProseMirror's mandatory initial NodeSelection as a blue ring to readers.
- **500px dead zone**: an invisible click-to-edit button below every document shifted the footer on hydration for readers who can't edit anyway.
- **Client-side pop-in**: comment counts, People tab, collaborators, and card fallback images now prefetch server-side — fewer client fetches *and* stable first paint.
- **JS-driven layout fallbacks**: header height and the content-width bonus moved from `useMedia`/measured state to CSS, so server and first client paint agree by construction.

## Cost

**Net −194 lines of production code** (+274/−997 across modified files, +529 in new modules) while covering **all** block types — the old renderer faked a subset and punted on the rest. Plus 8 new headless unit tests and the parity harness. `ssr-editor-rework-notes.md` documents the architecture, verification methodology, and known limitations (tweets are inherently client-only; math/table gaps noted).


horacioh:
The new renderer can emit real content-view embed markup, but the cache key still only includes the parent document version plus query-block results. If an embedded document changes while the parent document version stays the same, the cached parent SSR HTML continues serving the old embedded content. The hydrated editor will then render the fresh embedded resource from the prefetch cache, causing stale no-JS/SEO content and a visible SSR-to-client mismatch for that embed.

Failure scenario:
Document A contains a content embed of Document B. A is requested once, caching SSR HTML. B is edited or republished, but A is not changed. A later requests hit the HTML cache at this key and render B's old content server-side, while hydration shows B's new content.

we should include the rendered dependencies for embeds in the cache key, for example the prefetched embedded resource versions for refs that can affect SSR output, or disable HTML caching for documents with content embeds until dependency fingerprinting is added. This might not be fully correct because this will not consider query blocks items either. so I'm not really sure what can we do here.

horacioh:
The generic React node-view SSR path now includes the math block, but the math component populates KaTeX output in a React effect. Effects do not run during renderToString, so SSR emits the math wrapper with an empty `<p>` instead of the formula. The old string renderer at least rendered the math text, so documents with math now lose server-rendered content until the client mounts.

Failure scenario:
A public document contains a math block like `E = mc^2`. With JavaScript disabled, or before hydration, the math block appears empty in the server-rendered document. On hydration, the effect runs and the formula appears, creating a content flash and hurting no-JS/SEO output.

I guess we should also server-render the Katex renderer?

ericvicenti:
I'm not going to worry about the math block for SSR because its a pretty niche feature. But I will test client rendering of math blocks before merging.

As for the loading concerns, I'm thinking about it and would like to discuss with you at your earliest convenience. 

When you refer to th HTML cache, are you talking about the current behavior, or something in the future? I don't think we do any HTML caching right now

ericvicenti:
From Claude Fable 5:

@horacioh — the embed-staleness concern was valid, and it's now fixed structurally in 11ddf88: the cache key fingerprints the **entire prefetched dataset** (every dehydrated `[queryHash, data]` pair, sorted, volatile timestamps excluded) instead of just query-block results. Since the fingerprint is computed from data fetched fresh on every request, any change in *anything* the render reads — embedded docs, query results, comment counts, account metadata — misses the cache and re-renders. A hit can never serve stale markup, and nothing needs to be remembered when future renders read new data.

For context on why we kept the cache rather than deleting it, we measured its value on real pages (some history: the cache dates to #459, where it shipped as an implementation detail of the original string renderer):

| Page shape | Uncached render | HTML |
|---|---|---|
| Real 190-block doc, text + highlighted code | 14.7ms | 81KB |
| Query List, 30 items (/notes-like) | 9.9ms | 73KB |
| Query Card grid, 30 items (home-like) | 14.6ms | 89KB |
| Real /design doc, 6 content embeds | 3.8ms | 20KB |

TTFB is ~170–300ms, dominated by the prefetch waves that run fresh per request regardless — so worst case a miss costs ~5–8% of TTFB, and with the full fingerprint a miss only happens when the data actually changed. The bigger page-speed lever (payload-level stale-while-revalidate with subscription-driven invalidation) is a separate design discussion.

On math SSR: agreed it's out of scope here; noted as a known limitation in the PR notes.

PR #858, opened 2026-07-09
fix(documents): cache hydrated documents by version to end read-path CPU storm

ericvicenti:
## Summary

Production hyper.media was intermittently unavailable (requests timing out at 20–60s, flapping in and out). Root cause: `Document.Hydrate` replays a document's **entire block-move op-log** on every read to rebuild tree state (`treeOpSet.State` → `isAncestor`), which is **superlinear in edit history**. Under load, repeated expensive hydrations of the same hot documents blocked each other and grew an unbounded request queue until the daemon collapsed.

This PR adds a **version-keyed hydrate cache + singleflight** in front of the two hot read paths (`GetDocument`, `getResource`). A resolved document version is content-addressed and immutable, so the hydrated proto for a given `(iri, version)` never changes and is safe to cache indefinitely. `singleflight` collapses concurrent identical hydrations into a single computation.

## Evidence

**Root cause reproduced locally** — `Hydrate` is O(n²) in tree depth (500 moves → 12ms, 1k → 47ms, 2k → 205ms, 4k → 873ms), with a CPU profile byte-identical to a production profile (95% of daemon CPU in `isAncestor`/`State`/`Hydrate`).

**Production telemetry** (`/debug/journeys`): ~50% of GetDocument load is version-pinned (immutable) and ~60% concentrates in just 26 documents re-hydrated ~495×/hour each — all redundant re-computation.

**Benchmark:** cache hit ~0.47ms vs ~884ms uncached for a 4000-move deep doc (~1,880×).

**Deployed to prod as a hotfix** (side-loaded image, watchtower paused) and verified:

| Metric | Before | After |
|---|---|---|
| `hyper.media` home | timeout 20–30s | 0.9–1.7s, stable 200 |
| `design/stories/authoring` | 78s | 0.05s |
| `debate-tema-7` doc | 16.7s | 0.009s |
| 8 concurrent reads | piled to 60s+ | all <90ms |
| Flaps in 10+ min | constant | zero |

## Correctness

The cache is **self-invalidating**: a `latest` read always resolves current heads via `loadDocument`, so after an edit it forms a new key and recomputes; old version-pinned entries stay valid because the content is immutable. Returned protos are defensive clones so shared cache entries are never mutated by callers. The visibility/deny check runs before the cache, so access control is unchanged.

New tests (`hydrate_cache_test.go`): cached output equals direct hydrate; independent clones; keys on version (invalidates on edit); concurrent singleflight correctness. All pass under `-race`. Benchmarks in `hydrate_bench_test.go` document the O(n²) scaling and the cache speedup.

## Files
- `hydrate_cache.go` (new) — version-keyed LRU (2048) + singleflight
- `documents.go`, `resources.go` — wire the cache into `GetDocument` and `getResource`
- `docmodel/docmodel.go` — export `Document.Version()` (content-addressed cache key)
- `*_test.go` (new) — correctness + benchmarks

## Deploy / durability

The fix is currently **side-loaded** onto the prod box (`docker save | ssh | docker load`) with **watchtower paused** so it can't revert to the old registry image. It survives reboots (compose uses the local `:latest` tag) but nothing auto-updates while watchtower is down.

**To make it durable:** `seedhypermedia/site:latest` is published only by a version tag (`*.*.*`) or a manual run of `release-docker-images.yml` from `main` — **not** by merging this PR. After merge, publish `:latest` (tag a release or dispatch the release workflow from main), then re-enable watchtower (`docker start autoupdater` on the prod box) so it picks up the registry image.

- Cap concurrent miss-hydrations with a semaphore for CPU headroom.
- Make `isAncestor` use O(1) map lookups instead of the btree hot path, so even cache misses are cheap.

### DIFF
--- a/backend/api/documents/v3alpha/docmodel/docmodel.go
+++ b/backend/api/documents/v3alpha/docmodel/docmodel.go
@@ -535,6 +535,14 @@ func (dm *Document) Heads() map[cid.Cid]struct{} {
 	return dm.crdt.Heads()
 }
 
+// Version returns the content-addressed version (canonical sorted heads) of
+// the document in its current state. It is deterministic and immutable for a
+// given set of applied changes, which makes it a safe key for caching the
+// hydrated result.
+func (dm *Document) Version() Version {
+	return dm.crdt.Version()
+}
+
 // Hydrate hydrates a document.
 func (dm *Document) Hydrate(ctx context.Context) (*documents.Document, error) {
 	if len(dm.crdt.changes) == 0 {
--- a/backend/api/documents/v3alpha/docmodel/hydrate_bench_test.go
+++ b/backend/api/documents/v3alpha/docmodel/hydrate_bench_test.go
@@ -3,6 +3,7 @@ package docmodel
 import (
 	"context"
 	"fmt"
+	"os"
 	"seed/backend/blob"
 	"seed/backend/core/coretest"
 	"seed/backend/util/cclock"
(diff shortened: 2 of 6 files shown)


ericvicenti:
Merging because @burdiyan says its okay, and because it seems to improve our production CPU bottleneck

PR #868, opened 2026-07-13
fix(activity): stop ListEvents pagination from silently skipping feed windows

### LINKED ISSUE #863: New comments are not showing up in the Activity Feed
Bea created a comment today at 15:17 CET. However, I don't see the comment in the activity feed:

https://seedteamtalks.hyper.media/design/stories/settings/as-a-user-i-want-to-manage-my-account-on-the-desktop-app/:comments/z6MkkGNjfnikRkxK9EaxMPYy3HYxNGkRyXTWxxRtynB2kyeb/z6HeLxpGK5Kst5

Image

iris-intelligence:
Fixes #863 — *New comments are not showing up in the Activity Feed.*

## TL;DR

- **Symptom** — comments (and doc updates, contacts, citations…) randomly vanish from the Activity Feed, even though they're fully indexed and visible in the document's Comments tab.
- **Cause** — `ListEvents` merges two independently-limited queries (blobs + mentions) and pages on the *minimum cursor of the returned page*. When an event is dropped from a page by the deleted/dedup filter, an **old mention slips in and poisons the next-page token**, silently skipping every event in between — hours of feed history at a time.
- **Fix** — clamp each page to the depth both queries actually scanned, so the token can never jump past unscanned rows. Two adjacent pagination dead-ends fixed along the way. Reproduced live on `seedteamtalks.hyper.media` and locally with a failing test before the fix.

## The symptom

Bea commented at 15:17 CET on the *"manage my account on the desktop app"* design story. The comment is on the server — it renders in the Comments tab and in a `pageSize=100` feed query — but the Activity Feed at `/:feed` never shows it. In the issue screenshot the feed jumps straight from **17:37 today** to **Jul 12, 11:27**, swallowing ~35 events from every author and event type in between.

## Root cause

`ListEvents` (`backend/api/activity/v1alpha/activity.go`) builds a page like this:

1. Fetch up to `PageSize` newest **blob events** (comments, doc updates, …) — `ORDER BY ts DESC LIMIT PageSize`.
2. Fetch up to `PageSize` newest **mention events** (citations, embeds) — same limit, **independent scan**.
3. Merge, drop deleted/moved/duplicate events, sort by event time, truncate to `PageSize`.
4. `nextPageToken = min(cursor in returned page) − 1`.

The two scans reach very different depths in time. The web UI requests `pageSize=5` with a link-type filter that matches few mentions, so the blobs scan covers the last ~20 minutes while the mentions scan reaches back **days**. That's fine while truncation cuts the old mentions off the page — but the moment step 3 drops a blob event (tombstoned comment, moved resource, dedup), an old mention rises into the page:

```
blobs scan (LIMIT 5):     15:56  15:51  15:50  15:37  [15:27 ← dropped by filter]
mentions scan (LIMIT 5):  Jul-12 11:27  …                    (sparse, reaches days back)

page returned:            15:56  15:51  15:50  15:37  Jul-12 11:27
nextPageToken:            min(page) − 1  =  Jul-12 11:26         ⟵ poisoned
                                    │
              everything in (Jul-12 11:27 … 15:27] is *never emitted on any page*
```

The blobs scan only covered down to 15:27; the token jumps to Jul 12. Every event in the gap — Bea's comment included — is unreachable forever. The window shifts as new events arrive, which is why the feed "randomly" loses different stretches of history.

## Smoking gun

Replaying the exact frontend request (`pageSize=5`, the UI's `filterEventType` list, site `filterResource`) against the live server and following `nextPageToken`:

```
page 0:  doc-update 16:03 · comment 15:56 · comment 15:51 · doc-update 15:50 · citation 10:06
         next token → 10:06:29.843                       ⟵ jumped past 5½ hours
page 1:  citation 10:06:26 · comment 09:52 · …           ⟵ resumes below the gap
…
15 pages walked: Bea's 13:17 UTC comment never appears.
```

The same query with `pageSize=100` returns the comment at its correct position — the data was always there; pagination was skipping over it.

## The fix

Track each query's **scan floor** (the smallest cursor it returned). A query that filled its `LIMIT` only covered rows down to its floor, so:

1. **Coverage clamp** — before truncation, drop merged events below the deepest floor among the fetches that filled their limit. They fall inside territory the other query hasn't scanned yet and will be re-fetched by the next page. The token now can never skip past unscanned rows.
2. **No empty-page dead-end** — if filtering empties a page while rows remain, emit a token at the coverage boundary instead of terminating the feed.
3. **Truncation counts as "has more"** — when both fetches come back under-full but the merged list still exceeds `PageSize` (feed tail), the truncated remainder previously vanished with no token; now it's reachable.

## Verification

- **`TestListEventsPaginationDoesNotSkipWindowWhenOldMentionEntersPage`** — recreates the incident against the real SQLite schema (5 fresh blob events, one dedup-dropped, one ancient mention) and pages exactly like the frontend. **Fails before the fix** (`[]int64{1, 3, 4, 5} does not contain 6`), passes after.
- **`TestListEventsEmitsNextTokenWhenTruncationCutsUnderFullFetches`** — covers dead-end (3); confirmed to fail with the new condition neutralized.
- Full `./backend/api/...` suite passes; `go vet` / `gofmt` clean. Applies identically to claimed-time and observed-time ordering (cursors are `ts` / blob id respectively).

**Follow-up worth considering:** the frontend defaults to `pageSize: 5` in `listEventsImpl`, which makes pages tiny and maximizes exposure to this class of bug (any single dropped event lets a stale mention into the page). The backend is now correct at any page size, but raising that default would cut round-trips and add margin.

### DIFF
--- a/backend/api/activity/v1alpha/activity.go
+++ b/backend/api/activity/v1alpha/activity.go
@@ -266,6 +266,11 @@ func (srv *Server) ListEvents(ctx context.Context, req *activity.ListEventsReque
 	// Count rows returned by the main DB fetch (pre-filter). Used to decide whether to
 	// emit a next-page token even if the deleted/dedup filters shorten the visible page.
 	var mainRawCount int
+	// Smallest cursor value returned by the main DB fetch (rows stream in
+	// cursor-descending order, so the last row sets it). Together with
+	// mentionsScanFloor this bounds how deep the merged page may reach — see
+	// the coverageFloor clamp below.
+	var mainScanFloor int64
 	if err := srv.db.WithSave(ctx, func(conn *sqlite.Conn) error {
 		err := sqlitex.ExecTransient(conn, getEventsStr, func(stmt *sqlite.Stmt) error {
 			mainRawCount++
@@ -318,6 +323,7 @@ func (srv *Server) ListEvents(ctx context.Context, req *activity.ListEventsReque
 			if orderByObserved {
 				cursor = id
 			}
+			mainScanFloor = cursor
 			eventCursors = append(eventCursors, feedCursor{
 				cursorValue: cursor,
 				blobID:      id,
@@ -407,6 +413,9 @@ func (srv *Server) ListEvents(ctx context.Context, req *activity.ListEventsReque
 	// Count rows returned by the mentions DB fetch (pre-filter). Used together with
 	// mainRawCount to decide whether more pages exist after post-fetch filtering.
 	var mentionsRawCount int
+	// Smallest cursor value returned by the mentions DB fetch (same contract
+	// as mainScanFloor).
+	var mentionsScanFloor int64
 
 	// Add mentions to the events list
 	if err := srv.db.WithSave(ctx, func(conn *sqlite.Conn) error {
@@ -526,6 +535,7 @@ func (srv *Server) ListEvents(ctx context.Context, req *activity.ListEventsReque
 			if orderByObserved {
 				cursor = blobID
 			}
+			mentionsScanFloor = cursor
 			eventCursors = append(eventCursors, feedCursor{
 				cursorValue: cursor,
 				blobID:      blobID,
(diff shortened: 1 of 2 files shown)


juligasa (approved):
I do think this will solve Bea's issue. I added some testing for coherence

PR #890, opened 2026-07-24
Add create site (space) flow for web and desktop apps

iskaktoltay:
Adds a "Create a Space" flow - a multi-step panel for creating a space, shared between desktop and web.

- UI: `CreateSpaceForm`, which collects name, cover, logo, and appearance settings,
   mapped to home-doc metadata by the shared `createSpaceMetadata`.
- Desktop: the sidebar "Create my Site" CTA opens the form in a drawer,
  writes a populated home-document draft under the selected identity, and
  opens the editor. Publishing the draft creates the space. The CTA is
  hidden once that identity already has a home document.
- Web (signed out): creates an anonymous `pending-` draft, lets the user
  edit it, and on publish creates an identity (vault round-trip), re-keys the
  draft to the new account, and publishes it.
- Web (signed in): creates a real draft directly and publishes.

### First-publish genesis bootstrap

A brand-new home document has no genesis on the server, and
`PrepareDocumentChange` can't bootstrap one. `publishWebDocument` now
publishes a client-signed genesis `Change` and version `Ref` first, then bases
the change on that genesis, mirroring the SDK's `publishDocument`
first-publish path.

## Testing

- `createSpaceMetadata` mapping (header-layout table, ipfs cover/icon) unit tests.
- Genesis first-publish (`publishWebDocument`) unit test (asserts genesis Change + Ref, then `PrepareDocumentChange` based on genesis).
- `CreateSpaceForm` step rendering unit tests.

## Known follow-ups

- **No web entry point yet**: `/hm/create-site` is reachable only by manually inputting the URL. The
  marketing/avatar/profile entry points are deferred until designs are ready.
- `site` to `space` rename and removing "document settings" from the
  3-dots menu are deferred to post-merge.

### DIFF
--- a/frontend/apps/desktop/src/components/create-space-dialog.tsx
+++ b/frontend/apps/desktop/src/components/create-space-dialog.tsx
@@ -1,0 +1,88 @@
+import {useSelectedAccountId} from '@/selected-account'
+import {client} from '@/trpc'
+import {fileUpload} from '@/utils/file-upload'
+import {useNavigate} from '@/utils/useNavigate'
+import {hmId} from '@shm/shared'
+import {Dialog, DialogSideContent, DialogTitle} from '@shm/ui/components/dialog'
+import {createSpaceMetadata} from '@shm/ui/create-space-platform'
+import {CreateSpaceForm, type CreateSpaceFormState} from '@shm/ui/create-space-form'
+import {Spinner} from '@shm/ui/spinner'
+import {SizableText} from '@shm/ui/text'
+import {toast} from '@shm/ui/toast'
+import {nanoid} from 'nanoid'
+import {useCallback, useState} from 'react'
+
+// Imperative opener for the "Create a space" panel.
+export function useCreateSpaceDialog() {
+  const [isOpen, setIsOpen] = useState(false)
+  const open = useCallback(() => setIsOpen(true), [])
+  const close = useCallback(() => setIsOpen(false), [])
+  const content = (
+    <Dialog open={isOpen} onOpenChange={setIsOpen}>
+      <DialogSideContent>
+        <DialogTitle className="sr-only">Create a space</DialogTitle>
+        {isOpen ? <CreateSpaceFlow onClose={close} /> : null}
+      </DialogSideContent>
+    </Dialog>
+  )
+  return {open, close, content}
+}
+
+/**
+ * The form with its desktop commit: populate a home document draft under
+ * the selected identity with the collected metadata, then open the home editor.
+ */
+export function CreateSpaceFlow({onClose}: {onClose: () => void}) {
+  const navigate = useNavigate()
+  const selectedAccountId = useSelectedAccountId()
+  const [busy, setBusy] = useState(false)
+
+  async function handleComplete(state: CreateSpaceFormState) {
+    if (!selectedAccountId) {
+      toast.error('Create or select an identity before creating a space.')
+      return
+    }
+    setBusy(true)
+    try {
+      // Upload images to IPFS, then write a populated home draft for
+      // the selected identity.
+      const [coverCid, logoCid] = await Promise.all([
+        state.cover ? fileUpload(state.cover) : Promise.resolve(undefined),
+        state.logo ? fileUpload(state.logo) : Promise.resolve(undefined),
+      ])
+      const metadata = createSpaceMetadata(state, {coverCid, logoCid})
+      await client.drafts.write.mutate({
+        id: nanoid(10),
+        editUid: selectedAccountId,
+        editPath: [],
+        metadata,
+        content: [],
+        deps: [],
+        visibility: 'PUBLIC',
+      })
+      // Open the home editor so the user can add content and publish.
+      navigate({key: 'document', id: hmId(selectedAccountId, {path: []})})
+      onClose()
+    } catch (e) {
+      toast.error('Failed to set up space: ' + (e instanceof Error ? e.message : String(e)))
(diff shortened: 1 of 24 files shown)


horacioh (changes requested):
# `feat/create-site` — issues found

Verified by code review plus end-to-end runtime testing on **desktop** (Electron + local daemon) and **web signed-out** (Remix + local dev vault), cross-checked against daemon gRPC metadata.

## Confirmed bugs

- **Uploaded "Space logo" is written to the wrong metadata field.** `createSpaceMetadata` maps the logo to `metadata.icon` and never writes `seedExperimentalLogo` (`frontend/packages/ui/src/create-space-platform.ts:21`). But `SiteLogo` renders the header logo *only* from `seedExperimentalLogo`, falling back to `HMIcon size={24}` when it's missing (`frontend/packages/ui/src/site-logo.tsx:19-40`). A 600×100 wordmark is therefore centre-cropped into a 24px circle — only the middle two letters survive. The Options panel confirms it: the image lands in **"Icon"** while **"Header Logo" stays empty**. The form previews the logo wide and uncropped, so the UI sets the wrong expectation. Affects **desktop and web**. Cover is fine on both.
  - *Decision needed:* the form collects one image, but the Options panel exposes both "Icon" and "Header Logo". Should the form write `seedExperimentalLogo` instead of, or in addition to, `icon`?

- **The "Set publication display date" toggle is a no-op.** `CreateSpaceFormState.showPublicationDate` is collected in the Appearance step but never mapped in `createSpaceMetadata`. There is no `showPublicationDate` field in `HMMetadata` — the nearest is `displayPublishTime`, which is a date string, not a boolean. Either wire it up or drop the control.

### Web-only — one shared root cause

The web create-space draft is addressed by the placeholder route `/hm/<uid>/-<draftId>` while its real publish target is the home doc (`editPath: []`, `frontend/apps/web/app/document-edit/web-create-space-draft.ts:58,70`). Everything downstream derives "is this a home doc?" from the **route path** rather than the draft's `editPath`, so the home draft is treated as a regular child document. Two visible symptoms:

- **The publish popover shows a phantom path, and silently ignores it.** `editing-toolbar.tsx:116-125` computes `isHomeDoc`/`isFirstPublish` from `docId.path`, so it renders a path preview plus an editable "Edit your permalink" input for a document that will have no path. Setting it to `/zzz-should-be-ignored` was accepted by the UI and updated the preview URL, but publish discarded it entirely — the doc landed at `/hm/<uid>` with `path: ""`, and the permalink renders "Document Not Found". `onPublishIntercept` takes no arguments, so the toolbar's `pathOverride` can never reach publish for space home drafts. **Desktop is unaffected** (it navigates straight to `hmId(uid, {path: []})` — bare UID, no path, no input).

- **The draft page renders the editable document header on a home document.** `EditableDocumentHeader` (the "Document Title" / "Document Summary" textareas) and the breadcrumb trail are gated behind `!isHomeDoc` (`resource-page-common.tsx:2212, 2274`, with `isHomeDoc = !docId.path?.length` at `1589`). The placeholder route defeats that gate, so the home-doc draft shows a title/summary editor and a `🏠 > QA Image Space` breadcrumb it shouldn't have.

- *Suggested fix for both:* derive `isHomeDoc` from the draft's `editPath` rather than the route path.

## Lower severity

- **Failed publish after account creation is silent.** In `pending-intent.ts:209-213`, the `publish-draft` branch catches the error, clears the intent, and returns `{type:'none'}` with no user-facing feedback. The user just created an account and would be left with no space and no explanation. Worth a toast.

- **Genesis bootstrap can leave an orphan blob.** In `publishWebDocument`, the genesis Change + Ref are published before the main change; if the subsequent `PrepareDocumentChange`/signed change fails, the genesis blob is already on the server. This mirrors the SDK's first-publish path, so probably acceptable — just noting it.

- **`SPACE_NAME_MAX_LENGTH = 20`** is quite short for a space name — confirming that's intended.

## Verified working

Name, `theme.headerLayout` (Center), `contentWidth` (S) and `showActivity` (false) all persist correctly to the daemon on both platforms; the cover renders on both; and the full signed-out web flow through vault account creation, pending-draft re-keying, and publish works end to end.

## Test environment notes (not product bugs)

- Chrome for Testing 133 on the test box lacks Ed25519 WebCrypto by default, so `chrome://flags/#enable-experimental-web-platform-features` was required. Default-on in Chrome 137+, so real users are unaffected. One real observation though: when Ed25519 is unavailable, "Create identity" fails **silently** — worth surfacing an error when session-key generation fails.
- Headless Chrome has no WebAuthn authenticator, so the password fallback was used; the passkey branch was not exercised.

image


https://github.com/user-attachments/assets/bfc11a38-44f6-4abe-b152-eb8c9c7e627c

iskaktoltay:
@horacioh Pushed a commit addressing the review. The form now writes seedExperimentalLogo. The avatar keeps its own field in the Options panel. Also removed set publication date toggle.

I checked the `editPath` to check if the document is home document, but I found that the draft-write actor rewrites `editPath` from the route on every autosave on web. From what I found, the only way to know if it's a home draft is the pending- uid when signed out, or a localStorage marker when signed in. So I centralized it into a small HomeDraftProvider context.

I also added a retry intent if the first publish fails after signing in. Now it re-keys the draft to a correct account UID and redirects there so the user can retry.

I also found that centered layout preview in the draft comes from the published home doc, not the draft, so right now the draft always shows horizontal layout until published.

horacioh (approved):
LGTM

PR #891, opened 2026-07-25
Fix: pooled-writer leaked transaction silently discards acknowledged writes (2026-07-24 prod data loss)

ericvicenti:
## Summary

Reproduces **and fixes** the mechanism behind the 2026-07-24 prod incident on hyper.media where **~2.5 hours of acknowledged writes (10:40–13:04 UTC) were silently rolled back** without any daemon restart.

The first commit adds failing tests that isolate the bug; the second commit fixes it and turns them green.

## The incident (evidence from prod)

- At `10:56:12Z` the agents service published a document (genesis + change + ref). The daemon logged `StoreBlobsWrite … blobCount:3 … Finished` (0.25s, success) and returned the CIDs. An immediate `Resource` read said `not-found` — and still does.
- Prod `db.sqlite` today contains **zero blobs with `insert_time` between 10:37:15 and 13:04:05**, despite successful `StoreBlobsWrite` log entries at 10:56, 12:00, and 12:08. The daemon did **not** restart between 09:11 and 23:41.
- Blob IDs were **reused**: the daemon logged `PutBlockExistsSample blobsID:120410/120411` at 10:58, but those IDs now hold *different* blobs inserted at 17:03 — the ID counter rewound when everything after ID 120393 rolled back. The two bitswap blobs from 10:58 were re-delivered by peers at 13:04:05 and became rows 120394/120395; the agent's document was published once and was never re-sent.
- `RBSRIndexServeFallback: "scopes lost materialization mid-load"` fired 20–100×/min from ~10:40 until exactly 13:04 — `materializeScope`'s work on the writer connection never became visible to read connections, so every reconcile re-materialized and warned. The warnings and the write blackhole end at the same minute.

## Mechanism

Three behaviors composed into an acknowledged-write blackhole on the pool's single write connection:

1. **`sqlitex.WithTx` error path** (`tx.go`): if the caller's ctx is canceled mid-transaction, the binding fails every subsequent call on the tripped interrupt — including WithTx's own `ROLLBACK`. WithTx returned an error, but the connection was left **inside the open transaction**.
2. **`Pool.Put`** (`pool.go`): checked for unreset statements (`checkReset`) but never `sqlite3_get_autocommit()`, so the poisoned connection re-entered the pool with its transaction open.
3. **`WithTx` nested fallback** (`tx.go`): every later writer leasing that connection hit `"transaction within a transaction"` and silently became a **savepoint** inside the leaked transaction. It reported success; its data was invisible to read connections; and it was destroyed wholesale when the leaked transaction eventually rolled back.

## The fix

Both ends are closed, so neither depends on the other:

- **`WithTx` repairs its own failures** (`tx.go`): a new `forceRollback` masks the interrupt (`SetInterrupt(nil)`, restore after) and rolls back whenever the plain `ROLLBACK` fails or a `COMMIT` fails without unwinding the transaction. Masking is safe: `ROLLBACK` only unwinds state the connection already owns and cannot block on the writer lock. When the recovery succeeds, WithTx returns the body's original error instead of a `ROLLBACK error`.
- **`Pool.Put` is the safety net** (`pool.go`): any connection returned inside an open transaction — from *any* code path, e.g. a top-level `Save` whose release was interrupted — is rolled back before re-entering the pool. Repair rather than panic, because the common trigger (a canceled request ctx) is normal operation. Repairs are counted in a new `sqlitex.LeakedTxRepairs()` counter and surfaced in red on `/debug/sqlite`, so a leaking code path stays visible instead of silently absorbed.

An uncommitted leaked transaction is rolled back, never committed on the leaker's behalf — the leaker was already told its transaction failed.

## Tests

`backend/util/sqlite/sqlitex/tx_leak_test.go` — public API only:

- `TestWithTxInterruptedRollbackLeavesAutocommit` — pins the source fix (fn-error path): WithTx must never return with the transaction open, and surfaces the body's error.
- `TestWithTxInterruptedCommitRollsBack` — pins the commit path: an interrupted COMMIT is reported as failure and the transaction is rolled back, not left open.
- `TestPoolPutRepairsLeakedOpenTransaction` — pins the safety net independently (raw `BEGIN IMMEDIATE` leak): repaired, counted, and the uncommitted write discarded.
- `TestPoolPutMustNotRecycleOpenTransaction`, `TestWithTxAcknowledgedWriteVisibleToReaders`, `TestWithTxAcknowledgedWriteSurvivesLeakedTxRollback` — the incident contract: after an interrupted request, leased connections are clean, acknowledged writes are reader-visible, and they survive later rollbacks on the pooled connection.

`backend/api/daemon/v1alpha/daemon_writer_tx_leak_test.go` — the incident shape at the gRPC surface:

- `TestStoreBlobsAckMustBeDurable` — after an interrupted request, `StoreBlobs`'s acknowledgment must be truthful: the blob is readable via `Index.Has` on a read connection. (Before the fix this test also took 10s instead of 0.01s — the read stalled behind the leaked writer lock.)

All of the above were red before the fix commit and are green after it. Full `./backend/util/sqlite/...`, `./backend/blob`, and `./backend/storage/...` suites pass.

## Side discovery: debug-tracker label budget

Adding these tests initially broke five unrelated `/debug/sqlite` page tests in the full-suite run only. Root cause: `maxCallerLabels = 64` — the Prometheus-cardinality guard — is a budget the test suite itself spends from (each test function that opens an instrumented tx is a distinct label), and the suite already sat within a handful of labels of the cap. Any few added tests silently bucketed later tests' callers into `"other"`. Bumped to 128 (~8 MiB worst-case reservoir memory) with a comment documenting the coupling.

## Notes

- A separate issue surfaced during the same investigation: the daemon crash-looped at 23:41Z on `'annotations[0].starts': source data must be an array or slice, got int` while decoding a change blob — a malformed-blob DoS vector. Not addressed here.
- This mechanism is also a plausible root cause for historically "flaky" pushes to shared servers (pushes ACK'd, then silently discarded).

https://claude.ai/code/session_01G9GyuSrhLz4z7DPQ8scZuz

### DIFF
--- a/backend/api/daemon/v1alpha/daemon_writer_tx_leak_test.go
+++ b/backend/api/daemon/v1alpha/daemon_writer_tx_leak_test.go
@@ -27,12 +27,13 @@ import (
 // destroyed ~2.5 hours later when the leaked transaction rolled back — no
 // daemon restart involved.
 //
-// The leak sequence uses only public API: a WithTx whose request ctx is
-// canceled mid-transaction. WithTx's internal ROLLBACK fails on the tripped
-// interrupt, and Pool.Put recycles the connection without noticing the open
-// transaction. Every later StoreBlobs then nests as a silent savepoint inside
-// it. See the companion tests in util/sqlite/sqlitex/tx_leak_test.go for the
-// pool-level mechanics.
+// The trigger uses only public API: a WithTx whose request ctx is canceled
+// mid-transaction. Before the fix, WithTx's internal ROLLBACK failed on the
+// tripped interrupt and Pool.Put recycled the connection without noticing the
+// open transaction; every later StoreBlobs then nested as a silent savepoint
+// inside it. The fix (WithTx force-rollback + Pool.Put repair) must keep
+// StoreBlobs' acknowledgment truthful. See the companion tests in
+// util/sqlite/sqlitex/tx_leak_test.go for the pool-level mechanics.
 func TestStoreBlobsAckMustBeDurable(t *testing.T) {
 	u := coretest.NewTester("alice")
 	keyMaterial := []byte("0123456789abcdef0123456789abcdef")
@@ -53,9 +54,9 @@ func TestStoreBlobsAckMustBeDurable(t *testing.T) {
 
 	srv := NewServer(store, &mockedP2PNode{}, idx, tMgr, zap.NewNop())
 
-	// Leak an open transaction into the pool's write connection: an earlier
-	// request is canceled mid-transaction, so WithTx's ROLLBACK is interrupted
-	// and the connection returns to the pool still inside the transaction.
+	// The prod trigger: an earlier request is canceled mid-transaction, so
+	// WithTx's plain ROLLBACK is interrupted. Before the fix this leaked the
+	// open transaction into the pool's write connection.
 	poisonCtx, cancel := context.WithCancel(context.Background())
 	conn, release, err := store.DB().WriteConn(poisonCtx)
 	require.NoError(t, err)
--- a/backend/api/daemon/v1alpha/daemon_writer_tx_leak_test.go
+++ b/backend/api/daemon/v1alpha/daemon_writer_tx_leak_test.go
@@ -1,0 +1,93 @@
+package daemon
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"seed/backend/blob"
+	"seed/backend/core/coretest"
+	taskmanager "seed/backend/daemon/taskmanager"
+	daemon "seed/backend/genproto/daemon/v1alpha"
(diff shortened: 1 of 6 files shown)


ericvicenti:
@juligasa @burdiyan this seems very urgent

I have experienced data loss, [a document](https://hyper.media/hm/z6MkwC95WHU2kPECQv2VuAfNmzBybBg3H8MpT4Ly5gCYKLo4/unsplash-wallpaper-ipfs-demo) was written by my agent, success was reported by the daemon. But the data did not survive.

Also @GaboHBeaumont experienced data loss (he reported that comments disappeared)

This PR is the result of Fable diagnosing, testing and "solving" the issue. But I am not familiar with this code, so it is difficult for me to review

PR #900, opened 2026-07-28
perf(directory): derive fallback cover image at index time to kill per-card N+1

ericvicenti:
Directory pages were the slowest part of the web app (audit: 2.57s FCP on `/projects`, driven by 143 per-card document fetches). This PR makes listing cards fully self-sufficient: everything a card renders now arrives with the listing response itself, and **rendering a listing never fetches a document**.

### 1. Indexer-derived fallback cover (`DocumentInfo.first_image_in_content`)

Cards with no explicit cover/icon used to fetch the child's **entire document** just to find its first image block. The indexer now derives the first content image (in reading order) once per state-advancing Ref and exposes it as a typed field on `DocumentInfo` (proto3 `optional string`, field 15):

- **Injected deriver** — `blob` can't import `docmodel` (cycle), so `documents` implements the derivation as a pure function over raw change bytes; the daemon wires it into the Index *before* the backfill-reindex task can start (guarded by `hookMu`; `NewServer` re-wires for tests/embedders). It rides on `indexOpts` with the other per-call knobs.
- **Correct under CRDT merge** — derives from the generation's *merged* heads (what the read path renders), gathered on the indexing conn so in-transaction writes are visible. Only Refs that actually applied new changes derive; duplicates and the unstash cascade skip the O(history) replay.
- **Never stale** — the result is always written to the internal `$db.firstImageInContent` LWW attr (stripped from public metadata): empty string means "derived, no image", timestamped at `max(refTime, LastAliveRefTime)` so a late-arriving older Ref that completes a merge still wins. Presence distinguishes "not derived yet" (render coverless until reindex) from "known imageless".
- **Backfill** via a storage migration that schedules a full reindex (~3min on a 70k-blob data dir, measured).
- Frontend: cards use the field (explicit icon suppresses the fallback cover, matching prior precedence); the old `needsContentFetch`/`lazyResource` machinery and the SSR fallback-image resource prefetch are **deleted**.

### 2. Interaction summaries ride the listing (`ActivitySummary.children_count`)

Cards/list items showed comment + subdocument counts via a per-item `InteractionSummary` — 4 gRPC calls each (`ListCitations` + `GetDocument` + `ListDirectory` + `ListDocumentChanges`), prefetched for the first 30 items of every page render. Now:

- `ActivitySummary` carries `children_count` (alive direct children), computed inside the shared listing query as one index-seeking prefix-range subquery — no extra queries, uniform across all four listing RPCs.
- The query-block payload's per-item summaries are `{comments, children}` straight from listing fields; the per-item SSR prefetch is deleted (embed cards keep theirs — they render from a document ref, not a listing item).

## Measured (A/B vs main, same real 70k-blob data clone, warmed dev servers)

**`/projects` (163 cards)**

| | main | this PR | Δ |
|---|---|---|---|
| Browser `/api/Resource` per-card fetches | **118** | **3** | −97% |
| Server-side gRPC calls | 358 | 108 | **−70%** |
| TTFB (SSR) | 984ms | 720ms | −27% |
| load event | 1649ms | 1345ms | −18% |

**`/tech` (82 cards)**

| | main | this PR | Δ |
|---|---|---|---|
| Browser `/api/Resource` per-card fetches | 45 | 2 | −96% |
| Server-side gRPC calls | 279 | 86 | **−69%** |
| TTFB (SSR) | 861ms | 436ms | −49% |
| load event | 1456ms | 1094ms | −25% |

Prod impact should exceed the dev-mode timing deltas: the eliminated browser fetches are post-hydration waterfalls over the public internet (the direct cause of the audit's 2.57s FCP), and the ~3.3× server-side reduction compounds under concurrent traffic since directory pages have no HTTP caching.

Remaining known hotspot (out of scope, documented): `GetAccount` ×70 for 37 unique authors per page — request-scoped dedup or `BatchGetAccounts` is the natural follow-up.

## Verification

- Hardened via a 17-agent adversarial review (all findings fixed: stale-value LWW, merged-heads derivation, icon precedence, cleared-cover gating, startup ordering race) and end-to-end runs against a clone of a real data dir: reindex clean, derived values match actual content 8/8, imageless sentinel 10/10, listing metadata carries only authored keys.
- New tests: `TestFirstContentImage` (reading order, creation-order independence, empty links), `TestListDirectoryDerivesFallbackCoverImage` (derive → skip-on-cover → sentinel → clear-on-image-removal via `ListDirectory`), `TestListDirectoryChildrenCount` (direct children only).
- Full `go test ./backend/...` green; workspace typecheck + prettier green.

Design doc with the full architecture investigation: `notes/directory-cover-fallback-design.md`.


ericvicenti:
## Adversarial review + real-data verification (933d62220)

Ran a 17-agent review workflow (5 lenses × adversarial verification, 0 findings rejected) **and** an end-to-end test of the migration + backfill reindex against an APFS clone of a real desktop data dir (70,028 blobs, 5,261 document generations), with the daemon fully isolated from the network (testnet protocol ID + dead bootstrap peer + localhost listen + sync disabled).

### Confirmed findings → fixes

| Finding | Severity | Fix |
|---|---|---|
| Derived value never cleared: an image removed from the doc kept showing as the card cover (reproduced on real data: **11 of 1,549** derived values were stale) | major | Always write the derivation result; `""` is a sentinel meaning "derived, no image", LWW-timestamped at `max(refTime, LastAliveRefTime)` so late-arriving older Refs that complete a merge still win |
| Derivation used the incoming Ref's heads, not the generation's merged heads (what the read path renders) | minor | Derive from `dg.Heads` via new `changesFromHeadIDsConn` |
| Full O(history) replay ran per Ref inside the foreground write tx, incl. duplicate Refs and the unstash cascade | minor | Gate on `appliedNewChanges` — only state-advancing Refs derive. The once-per-write replay remains (accepted M3 decision) |
| `hasCover`/`hasIcon` presence checks treated a cleared (`""`/nil) cover as present, permanently disabling derivation | minor | `hasNonEmptyAttr` treats nil/empty as absent |
| Indexed fallback cover could override an explicit icon on cards (pre-branch, icons suppressed content-derived covers) | major | Icon gates `indexedFirstImage` in `newspaper.tsx` |

Bonus from the sentinel: imageless docs (**3,190** generations on real data) now ship `"" ` and their cards skip the fallback document fetch entirely — the residual N+1 for text-only docs is gone too.

### Real-data results (fixed build, fresh clone)

- Migration `2026-07-28.104351` fired, full reindex completed in **3m08s**, zero derivation errors.
- Derived-table counts identical to baseline (`structural_blobs`, `resources`, `document_generations`).
- **8/8** non-empty derived values match the document's actual first image block (verified via `GetDocument` content walk); the previously stale doc now correctly stores the empty sentinel.
- **10/10** sampled sentinel docs genuinely have no image.
- `ListDirectory` over gRPC serves `_firstImageInContent` alongside `name` in card metadata.

Note for reviewers: the sentinel appears in listing metadata for every cover-less/icon-less doc (`"_firstImageInContent": ""`), which is why several listing tests' expectations were updated — `Document` metadata (CRDT) intentionally does not carry the derived key, only listings do.

ericvicenti:
## Derived cover moved to a typed sibling field (85584a5df)

Per discussion: the derived value no longer lives in the metadata map. It's now **`DocumentInfo.first_image_in_content`** (`optional string`, field 15).

- **Authored metadata stays pure** — no `""` sentinel on every listing item, no reserved `_`-prefixed key that could round-trip into authored changes. The listing tests reverted to clean metadata comparisons and now assert metadata purity explicitly.
- **Storage unchanged in spirit** — same LWW register, now under the internal `$db.firstImageInContent` attr (stripped by `PublicMap`, same precedent as `$db.redirect`), surfaced by `documentInfoFromRow`.
- **Tri-state survives via proto3 `optional`** — unset = not derived yet (client may fall back to fetching the doc), `""` = derived, no image (client skips the fetch).
- **Frontend** consumes `HMDocumentInfo.firstImageInContent`; `DocumentCardGrid` passes it to `DocumentCard` as a prop. `HMDocumentMetadataSchema` no longer mentions the key.

Re-verified against a fresh clone of the real 70k-blob data dir: reindex clean (3m22s, zero derivation errors), `ListDirectory` serves the sibling field (41/42 docs in the sample directory, 8 with images), listing metadata contains only authored keys, and storage carries zero old-style public keys.

ericvicenti:
## Simplification: fallback-cover fetch removed entirely (e7b48d59b)

Per discussion — the transition window (minutes of post-deploy reindexing, web+daemon shipped together) doesn't justify keeping the old fetch path around. Cards now get their fallback cover from exactly two sources: the indexed `first_image_in_content` field (listing callers) or content already in hand (entity-bearing callers). **A card never fetches a document, and neither does the SSR loader on its behalf.** Not-yet-derived docs render coverless until the backfill reindex completes.

Deleted:
- `needsContentFetch` / `lazyResource` machinery in `DocumentCard`
- the `fallbackImageIds` resource prefetch in the web SSR loader (predated the indexed field; was fetching up to 30 full documents per directory page)

Measured on a fresh `/projects` load (163 cards, real-data clone): server-side `GetResource` **36 → 9** (page doc, home, latest-version check, embeds — all legitimate), total gRPC **243 → 216**, covers render identically.

Remaining call volume is pre-existing and orthogonal to this PR, for future work: ~124 calls from the per-card `InteractionSummary` prefetch (4 gRPC calls × 30 cards + page — batching candidate) and 70 `GetAccount` for 37 unique authors (dedup/batch candidate).

ericvicenti:
## Commit 2 (c124d441f): children count rides on listings — per-card interaction summaries removed

Cards and list items showed comment + subdocument counts by requesting a full `InteractionSummary` per item, which costs **4 gRPC calls each** (`ListCitations` + `GetDocument` + `ListDirectory` + `ListDocumentChanges`) — prefetched for the first 30 items of every directory-page render (~124 calls per load).

- `ActivitySummary.children_count` (field 6): alive direct children, computed inside the shared listing query as one column — an index-seeking prefix-range subquery (`iri > base || '/' AND iri < base || '0'`, with a depth filter and latest-generation-alive check), used by all four listing RPCs uniformly. Comment counts were already present.
- The query-block payload's per-item summaries now carry `{comments, children}` from those fields, so nothing per-item is fetched. The SSR per-item summary prefetch is deleted; embed cards keep theirs (they render from a document ref, not a listing item).
- New `TestListDirectoryChildrenCount` covers direct-children semantics (grandchildren excluded).

**Measured** (fresh `/projects`, 163 cards, real-data clone): server-side gRPC **216 → 110** — the quartet collapsed (`ListCitations` 32→5, `ListDocumentChanges` 31→4, `GetDocument` 32→5, `ListDirectory` 34→7), zero new client-side requests, cards render identically.

Cumulative across the PR: **243 → 110** server-side for a fresh directory load, and the browser went from 143 per-card document fetches (original audit) to 0. The remaining big bucket is `GetAccount` ×70 for 37 unique authors (dedup/batch candidate, untouched here).

PR #904, opened 2026-07-28
Fix: query blocks show CommentCount=0 for moved/redirected documents

ericvicenti:
## Problem

All ~115 docs under [seedteamtalks.hyper.media/tech](https://seedteamtalks.hyper.media/tech) show **0 comments** in the directory/query-block listing, even though many have active discussions (the doc pages themselves show the comments fine).

## Root cause

A move is a Version Ref at the new path + a Redirect Ref at the old path. Comment blobs record the target **path as it was when the comment was written**, so the indexer credits `document_generations.comment_count` on the **old path's resource**. Listings (`baseListDocumentsQuery`) read only the listed resource's own row — which starts at 0 after a move. `ListComments` is redirect-aware (`redirectAncestorsCTE`), which is why doc pages disagreed with listings. Verified on the production gateway DB: the old `/tech-talks/...` rows hold the counts + redirects, the new `/tech/...` rows hold 0.

## Fix

Compute listing comment stats at query time from the indexed Comment blobs: dedup edits by TSID, drop tombstoned comments, and credit each comment to its target resource **plus every transitive redirect target**. This exactly mirrors what ListComments serves, independent of blob arrival order and of how clients key the move Refs (desktop merges the redirect into the existing generation row; the API default creates a fresh one — both occur in prod and both are handled). Activity-time sorting uses the redirect-aware timestamp too.

**Perf:** listings pay a flat ~30–50 ms aggregation on the prod-size DB (same cost band ListComments already pays per request; amortized over the page). Point lookups (`getDocumentInfo` — called in loops by `BatchGetDocumentInfo` and ListAccounts home-doc hydration) use a separate aggregation **seeded backwards from the one requested IRI** (~1–3 ms, fully indexed plans). A new partial expression index `document_generations_by_redirect` (schema.sql + migration `2026-07-28.193121`) makes redirect-chain walks indexed. **The migration only creates the index — no reindex needed**, since stats are derived at query time.

**Pagination fix:** activity-sorted cursors now carry the same `activity_time` value the query orders by (previously `LatestChangeTime`, which could skip rows at page boundaries whenever comments/refs were newer than the last change — pre-existing, amplified by redirect-aware activity).

## Review notes (side effects considered)

- **Republish redirects** (doc alive at old path + republished at new): both paths stay listed and both count the comments — matching ListComments for each path. Covered by `TestCommentCount_RepublishRedirect`. (An earlier draft moved counts instead of crediting them, which zeroed republished docs — caught against prod data.)
- **Delete/edit after a move** now affects counts correctly (the incremental counter applied deltas to the wrong resource and clamped at 0, so a post-move delete could never decrement). Covered in `TestCommentCount_AfterDocumentMove`.
- Counts now include comments whose target changes aren't indexed yet (the incremental path silently dropped them at `blob_comment.go` `cm.ID == 0`); this matches what ListComments shows.
- Private comments remain included in counts (unchanged behavior).
- Anyone can redirect their own path at someone else's doc and thereby add to its count — inherited from ListComments' existing redirect semantics, not new surface.
- `spaces.comment_count` (ListAccounts space totals) is still incrementally maintained and untouched; it can drift from blob-derived doc counts (pre-existing).
- Redirect cycles are bounded: out-degree 1 per resource + depth cap 16 (same guard as `redirectAncestorsCTE`).
- Old page tokens stay decodable; one transitional page after deploy may skip/duplicate a row (transient).
- `document_generations.comment_count`/`last_comment*` are still written by the indexer but no longer feed listings; nothing new should be built on them.

## Validation

- New tests: `TestCommentCount_AfterDocumentMove` (pre/post-move comments, multi-hop chain, delete-after-move, ListDirectory/ListDocuments/ListComments agreement), `TestCommentCount_RepublishRedirect`.
- Ran the exact generated SQL read-only against the production gateway DB (6.5 GB): **700 documents corrected** — 699 recovered undercounts (`/tech` 0→293, exactly matching ListComments), 1 fixed overcount (deleted comment the incremental counter failed to subtract). EXPLAIN-verified index usage.
- documents/storage/blob/entities/activity/daemon suites pass; migration snapshot test passes.


ericvicenti:
Added a perf follow-up commit after measuring against a copy of the production gateway DB (6.5 GB, ~11k generations, ~14k comments):

**Measured costs of the aggregation:**
- Listing endpoints (ListDirectory/ListDocuments/ListRootDocuments): the global aggregation adds a flat ~30–50 ms per request (it must touch every comment row once to dedup edits by TSID). That's amortized over the whole page and is the same cost band `ListComments` already pays per request today (its `redirectAncestorsCTE` also does a full `document_generations` pass), so listings stay within existing norms.
- `getDocumentInfo` would have paid the same flat cost **per call**, and it's called in loops — `BatchGetDocumentInfo` and ListAccounts' home-document hydration would have turned that into seconds per request.

**What the second commit does:**
- Point lookups get their own aggregation seeded backwards from the one requested IRI (mirroring `redirectAncestorsCTE`), touching only that document's redirect chain and comments: **~1–3 ms per call** with fully indexed plans.
- New partial expression index `document_generations_by_redirect` (schema.sql + migration `2026-07-28.193121`, no reindex) so redirect-chain walks are indexed lookups; the listing aggregation's `redirected` CTE was restructured to use it too.
- Verified query plans (EXPLAIN) and results on the prod DB copy — counts identical to the first commit, e.g. `/tech` = 293 matching ListComments exactly.

If listing latency ever matters more (comment volume grows linearly into the window cost), the next step would be seeding the listing aggregation from the directory's IRI filter as well — noted as future work, not needed at current scale.

PR #905, opened 2026-07-28
fix(web): say where config loading looked and why it failed

ericvicenti:
The startup error `Failed to load configuration. Set DATA_DIR/config.json or DATA_DIR/service-config.json` doesn't say **where** it looked. The path resolves against `DATA_DIR` when set, otherwise against the process working directory — so launching the web app from any cwd other than `frontend/apps/web` (repo root, an IDE task) fails with a message that looks wrong when `config.json` plainly exists next to the app.

The error now includes the resolved directory, how it was chosen (`DATA_DIR` vs cwd), and the underlying failure for each of the two config files:

```
Failed to load web app configuration from /Users/x/Code/Seed (process cwd; set DATA_DIR to override):
  config.json: ENOENT: no such file or directory, open '/Users/x/Code/Seed/config.json'
  service-config.json: ENOENT: ...
```

Also drops the three leftover `~~ process.env.DATA_DIR` debug prints.

No behavior change — same resolution order, same throw condition. Standalone against main (unrelated to #900; found while testing it).

### DIFF
--- a/frontend/apps/web/app/site-config.server.ts
+++ b/frontend/apps/web/app/site-config.server.ts
@@ -8,9 +8,6 @@ import {z} from 'zod'
 export const adminSecret = process.env.SERVICE_ADMIN_SECRET
 
 const webDataDir = process.env.DATA_DIR || process.cwd()
-console.log('~~ process.env.DATA_DIR DATA_DIR', process.env.DATA_DIR)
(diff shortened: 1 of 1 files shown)


ericvicenti:
Withdrawn — the confusion came from local testing side effects, not a real gap.

PR #909, opened 2026-07-29
fix(backend): oversized changes fail to apply instead of crashing the daemon

ericvicenti:
## Production is crashing on this today

`daemon_gateway` on prod panicked **7 times in the last 72 hours** with `panic: BUG: idx too big`, each one triggered by a plain **`GetDocument`** call (verified in container logs; the restart count matches). There is no gRPC recovery interceptor, so each panic kills the whole daemon.

The trigger is `hm://z6MkpsGurNh1FnR1ZhUpdxGkHv4LKcAeYYuCjXrhcjziUbWY/es` ("Legislacion Espanola", account "demo publish"; formerly at `/spain`): its 1.3MB change contains a single `MoveBlocks` op with **6,411 blocks**. The ApplyChange loop advances the op ID index quadratically per block (`idx += i`), so it crosses the packed 24-bit limit at block 5,791 and panics `newOpID`. Anyone opening that page takes down the gateway — a public URL that works as a DoS button. This is the same landmine #900's deriver stepped on during the backfill reindex (fixed at the deriver boundary in #908); this PR fixes the underlying apply path that prod is crashing through.

## The fix

Validate the op ID index (and timestamp, which panics identically in the same packed representation) on the **apply** path and return an error instead of panicking. Serving such a document now fails that one request; the daemon keeps running.

This covers the write path too, rather than leaving it panicking: authoring a change runs through this same apply loop (`CreateChange` → `applyChangeUnsafe` → `crdt.ApplyChange`, `docmodel.go:377,201`), so a local writer that somehow produced an oversized change now gets an *error* as well — a failed save rather than a crash. What does keep its panic is the block-tree mutation in `crdt_block_tree.go:384,409`, whose counter advances linearly and genuinely cannot reach the limit, so a panic there still means a real bug. (Thanks @juligasa — an earlier version of this description claimed the write path still panicked; it doesn't, and nothing should be built on that assumption.)

Covered by tests at both levels: `docmodel.ApplyChange` returns the error directly, and `DeriveFirstContentImage` (index path) surfaces it without panicking.

## Sequencing

This wants to be **widely deployed before #910**, and the point is stronger than "old nodes keep failing on old documents." Once #910 ships, an import can *create* content that pre-#910 gateways cannot read at all — so the previously-frozen documents should not be edited until #910 is out across the gateways. Until then, this PR is what keeps a node encountering one of them alive.

## Not in scope

The document itself stays unopenable here (error instead of crash). Making it actually load means renumbering op IDs — the write side counts linearly, the apply side quadratically, so ≥3-block move ops get different identities depending on the scheme, and historical refs resolve against the applied numbering. That's #910.

### DIFF
--- a/backend/api/documents/v3alpha/docmodel/crdt.go
+++ b/backend/api/documents/v3alpha/docmodel/crdt.go
@@ -411,6 +411,23 @@ func (e *docCRDT) ApplyChange(c cid.Cid, ch *blob.Change) error {
 		panic("BUG: actor wasn't derived when applying change")
 	}
 
+	if ts != math.MaxInt64 && ts >= maxTs {
+		return fmt.Errorf("change %s: timestamp %d overflows the maximum op ID timestamp %d", c, ts, maxTs)
+	}
+
+	// A change carrying enough ops can push the op ID index past what the
+	// packed opID representation can hold (the index advances quadratically
+	// within multi-block ops). newOpID panics on such values because today's
+	// write path can never produce them, but changes arriving from the network
+	// can — huge imported documents from older writers exist in the wild — and
+	// applying those must fail with an error instead of crashing the process.
+	checkedOpID := func(i int) (opID, error) {
+		if i > maxIdx {
+			return opID{}, fmt.Errorf("change %s: op ID index %d overflows the maximum %d", c, i, maxIdx)
+		}
+		return newOpID(ts, actorID, i), nil
+	}
+
 	idx := -1
 	for op, err := range ch.Ops() {
 		idx++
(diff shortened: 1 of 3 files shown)


juligasa:
The fix is right, and the apply path is the correct boundary for it — that's where untrusted network input lands. Confirmed all five `newOpID` call sites in the loop are covered, and that a mid-apply error can't leak partial state to callers: `loadDocument` breaks and returns nil (`documents.go:1826`) and `Checkout` returns the error (`docmodel.go:158`).

One small inaccuracy in the description, not the code: "The write path keeps its panic via `newOpID` — locally authored ops count linearly and can't legitimately reach the limit." Authoring a change runs through this same apply loop — `CreateChange` → `applyChangeUnsafe` → `crdt.ApplyChange` (`docmodel.go:377,201`) — so after this PR a local writer that somehow produced an oversized change gets an *error* too, not a panic. That's better behaviour (a failed save rather than a crash), just worth describing accurately so nobody later relies on "the write path still panics" as an invariant. What does keep its panic is the block-tree mutation in `crdt_block_tree.go:384,409`, whose counter advances linearly and genuinely can't reach the limit.

Also worth restating the sequencing point from #910 a bit more strongly when these go out: it isn't only that old nodes keep failing on old documents. Once #910 ships, an import can *create* content that pre-#910 gateways cannot read at all. So this one wants to be widely deployed first, and the previously-frozen documents shouldn't be edited until #910 is out across the gateways.

PR #910, opened 2026-07-29
fix(backend): widen the op ID index so large imported documents load

ericvicenti:
Stacked on #909 (error-instead-of-panic). That PR stops the crashes; this one makes the affected documents actually load.

## Why the cap exists and why it's safe to raise

The 2^24−1 index cap is a fossil: it enforced the packed 15-byte `EncodedOpID` layout (48-bit ts / 24-bit idx / 48-bit actor) that was removed in 2024's Breaking Change. Today `opID` is a plain struct in memory and refs serialize the index as a full uint64 — the cap is policy, not representation.

Since the apply loop advances the index quadratically inside multi-block ops, a single move op of ~5.8k blocks reached the old cap. That's real-world size: the "Código Civil" document under the demo-publish account carries a 6,411-block move op (index peaks ~20.5M) and is what the production gateway has been crashing on. Raising the cap to 2^31−1 allows ~65k blocks per move op — 10× the largest op seen — while keeping a bound on hostile changes. The struct field widens to int64 so the representation is no longer the constraint, and `decodeOpID` now bounds-checks incoming ref indexes instead of silently truncating through an int32 cast.

## Compatibility

- **No migration, no reindex, nothing persisted changes.** Op IDs are recomputed in memory on every load; the wire encoding is unchanged. Reverting simply makes over-cap documents error again.
- Documents that load today produce byte-identical op IDs — the numbering scheme is untouched, only the ceiling moves.
- Sequencing: nodes still on older builds keep failing on these documents (panic before #909, clean error after), so this should ride a release *after* #909 is widely deployed — especially before anyone edits the previously-frozen documents, since new changes building on high-index ops would be unappliable on old builds.

## Verification

Beyond unit tests (a 6,411-block op applies; a 70k-block op fails cleanly), I pulled the real document's change blobs from the production database and replayed them through this code: all changes apply, and the document hydrates with full content and metadata ("Real Decreto de 24 de julio de 1889 por el que se publica el Código Civil").

### DIFF
--- a/backend/api/documents/v3alpha/docmodel/crdt.go
+++ b/backend/api/documents/v3alpha/docmodel/crdt.go
@@ -25,7 +25,7 @@ import (
 
 type opID struct {
 	Ts    int64
-	Idx   int32
+	Idx   int64
 	Actor core.ActorID
 }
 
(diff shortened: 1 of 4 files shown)


juligasa:
Reviewed both PRs and pushed 6388b08 here with three small additions:

- **`decodeOpID` now bounds-checks `Ts` too.** It was still doing `Ts: int64(s[0])` straight from the wire, so a ref carrying a large `uint64` decodes to a negative timestamp that sorts before every real op in `opID.Compare`. Same class as the index check this PR adds, same function.
- **A table test for `decodeOpID`.** It had no test at all, and this PR adds two rejection paths to it.
- **A comment on the apply loop's `idx += i`.** That line reads as an off-by-one and invites a one-character "fix", but the indexes it produces are wire-visible: existing changes encode them in their refs, so `idx++` would renumber every op ID in existence and historical refs would stop resolving (`causality violation: ref op ... is not found`) on documents that load fine today. Worth stating explicitly right where the next person will read it.

**On the ceiling — I think 2^31−1 is right as written.** I went in expecting to argue for raising it, since it caps a single move op at ~65k blocks (only ~10× the Código Civil op) and these documents come from bulk import, the one path that legitimately emits a huge move op. But content transfer goes over bitswap, not the 8 MiB gRPC cap (`backend/hmnet/client.go:53` says so explicitly), and there's no application-level size limit on a Change blob — so this cap is the only application-level bound on how much work a single change can demand. Raising it later is a one-character change with the same compatibility story as this PR; lowering it later is not.

Everything else checks out: the arithmetic (2^24−1 crossed at block ~5,793; 6,411 blocks → 20,547,255; 2^31−1 crossed at block 65,536), the no-migration claim (`opID.Idx` never reaches SQLite — only `crdt.go`, `docmodel.go:459`, and the write counter in `crdt_block_tree.go` touch it), and the int32→int64 widening (`Compare` unaffected in range, `encodeOpID` already casts to `uint64`, wire form is a CBOR varint).

Two things I'd suggest as separate PRs rather than growing this one:

1. **A gRPC recovery interceptor.** #908 recovered one deriver and #909 removes one panic source, but the class — a panic reachable from network input taking down the daemon — is still open, and neither `backend/daemon/daemon.go:378` nor `backend/hmnet/hmnet.go:209` has a net. `go-grpc-middleware/v2` is already in `go.mod` as an indirect dep, so `v2/interceptors/recovery` needs no new dependency.
2. **Pin the write/apply numbering asymmetry.** The writer numbers moves linearly (`crdt_block_tree.go:384,409`) and encodes a self-ref as the bare linear index, while the applier renumbers quadratically and keeps the writer's `Idx`. That's benign today *only* because `cleanupPatch`'s batching (`docmodel.go:456-464`) means a self-ref always continues the current batch, so a batch never starts with one — and only a batch's first ref is encoded. Nothing states or tests that invariant, and it's what makes the quadratic numbering un-fixable without a version bump.

PR #932, opened 2026-08-05
Repair missing RBSR migration tables

horacioh:
## What changed

Adds a forward storage migration (`2026-08-05.103117`) that idempotently recreates the derived RBSR tables/index when an existing data directory is missing them:

- `rbsr_scope`
- `rbsr_item`
- `rbsr_item_by_blob`

## Why

A Windows user reported that the desktop app appeared to do nothing on launch/install. The pasted logs show the daemon does start, but then exits almost immediately:

```text
DaemonStarted
P2PNodeReady
GracefulShutdownStarted
sqlite.Exec: Conn.Prepare: SQLITE_ERROR: no such table: rbsr_item
Go daemon closed code=1
```

After the daemon exits, the desktop keeps polling gRPC and only sees `ECONNREFUSED`, eventually timing out with `Timed out: waiting for daemon gRPC to be ready`.

The root cause is an existing user data directory that had advanced beyond the original RBSR migration version while missing the `rbsr_*` derived tables. Reindex/sync code now expects those tables, so startup can fail before the desktop gets a stable daemon connection.

## Impact

Affected existing installs should repair themselves on next startup instead of failing daemon boot. Fresh data directories are unchanged because these tables already come from `schema.sql`. The tables are derived, so creating them empty is safe; they can re-materialize lazily as existing RBSR code runs.

## Validation

Ran:

```sh
go test ./backend/storage -run TestMigrationRestoresMissingRBSRTables -count=1
go test ./backend/storage -count=1
go test ./backend/blob ./backend/daemon ./backend/storage -count=1
go test ./backend/...
golangci-lint run --new-from-merge-base origin/main ./backend/...
```

All passed; lint reported `0 issues`.

### DIFF
--- a/backend/storage/storage_migrations.go
+++ b/backend/storage/storage_migrations.go
@@ -63,6 +63,27 @@ type migration struct {
 //
 // In case of even the most minor doubts, consult with the team before adding a new migration, and submit the code to review if needed.
 var migrations = []migration{
+	// Repair installs that already advanced past the original RBSR-table
+	// migration without receiving it. The tables are derived and safe to create
+	// empty; fresh data dirs already have them from schema.sql.
+	{Version: "2026-08-05.103117", Run: func(_ *Store, conn *sqlite.Conn) error {
+		return sqlitex.ExecScript(conn, sqlfmt(`
+			CREATE TABLE IF NOT EXISTS rbsr_scope (
+			    id INTEGER PRIMARY KEY,
+			    iri TEXT NOT NULL,
+			    kind INTEGER NOT NULL,
+			    materialized INTEGER NOT NULL DEFAULT 0,
+			    last_access INTEGER NOT NULL DEFAULT 0,
+			    UNIQUE (iri, kind)
+			);
+			CREATE TABLE IF NOT EXISTS rbsr_item (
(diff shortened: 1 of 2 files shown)


burdiyan:
This is super weird. I really don't get how on earth that would have happened:

> The root cause is an existing user data directory that had advanced beyond the original RBSR migration version while missing the rbsr_* derived tables. Reindex/sync code now expects those tables, so startup can fail before the desktop gets a stable daemon connection.

@horacioh do you know what the `<data-dir>/daemon/VERSION` file says on the user's machine?

horacioh:
@burdiyan 2026-07-28.193121

burdiyan:
Then I really don't understand how could he not have those `rbsr_*` tables. It's extremely weird. I'm concerned with adding band-aid migrations to re-create those tables without trying to understand why are they missing in the first place.

burdiyan:
https://github.com/seed-hypermedia/seed/blob/8bcabd706ac24c80df87c22d3b5d87a2640ea7b7/backend/storage/storage_migrations.go#L106-L109

@juligasa could this be related to what happened to Joel?

burdiyan:
I made some changes. LGTM.

@juligasa please have a final say on this, and merge if OK. I scheduled the reindex of the database, hope they won't take hours now.

juligasa (approved):
LGTM

PR #958, opened 2026-08-17
The agents UI moves to @shm/ui, and the web app gets it at /hm/agents

ericvicenti:
The agents UI moves out of the desktop app into `@shm/ui/agents`, and the web app mounts the same screens at `/hm/agents`.

This is the web-agents work from August, replayed onto current main. The original branch forked before The Harness (#955) landed, so this is not a mechanical rebase: every relocated file was three-way merged against main's current version, so the ~20 commits of agents work that landed after the fork are preserved rather than reverted to the fork-point copy.

## What's here

**The UI moves to the shared package.** 27 modules move from `frontend/apps/desktop/src/{pages/agents,components,models}` into `frontend/packages/ui/src/agents/`. Everything app-specific goes through one seam, `agents/platform.ts`: daemon signing vs. the web device key, trpc app settings vs. `localStorage`, the local HM node, the loopback OAuth catcher, and the rich editor (injected, because `@shm/editor` depends on `@shm/ui` and importing it directly would cycle).

**Importers point at the real location.** There are no re-export shims: desktop imports `@shm/ui/agents/*` directly, so each module has exactly one path and nothing to chase through.

**The web app gets the agents UI.** `/hm/agents` and its subroutes, a web platform adapter built on the local device key, and the agent server URL from `SEED_AGENT_SERVER_URL`. The agents bundle is code-split — it costs a 298 kB chunk only on the agents routes.

**Phones.** The composer uses 16px on small screens (iOS Safari zooms the page whenever a focused field is smaller), and touch targets grow.

## The one contract change

Main replaced the fire-and-forget `discoverReferences` with a live subscription (`8f3908222`, "sync content for open remote sessions") because a one-shot discover can race the peer connection or return a result cached from before the agent published. Keeping that behavior through the seam needed a new optional member, `subscribeToEntity`. Desktop implements it over `client.sync.subscribe`; the browser has no local node, so web omits it and the shared code no-ops.

Every optional seam member is guarded at the call site — that is what makes the same code run in a browser that cannot sync, connect to peers, or open a loopback socket.

## Verification

- `pnpm typecheck` clean across the workspace
- desktop 680 tests, `@shm/ui` 314, shared 1046, web 158 — all passing
- `pnpm format:check` clean
- the web app builds
- new suite for the web platform adapter (`web-agents-platform.test.ts`, 10 tests): settings round-trip through namespaced `localStorage`, SSR inertness, corrupted values, refusing to sign without an identity, and — the one that matters — asserting the adapter omits every capability the browser cannot honor. Break-tested: adding `subscribeToEntity` to the web adapter or changing the storage prefix fails it.

## Review findings, fixed here

A review pass diffed every relocated file directly against main and confirmed no main-side behavior was lost. Three findings came out of it, all addressed:

- **`@shm/ui` imported `remark-gfm-v4` without declaring it.** `markdown.tsx` moved into the package still importing it, but only the desktop app declared it, so it resolved through hoisting rather than by contract. That defeats the point of the alias — a bare `remark-gfm` is ambiguous here (the editor's v3 wins the renderer prebundle and `react-markdown@10` throws on it), which is why the pin exists. Desktop no longer imports it, so the declaration moved rather than being duplicated.
- **Agent `hm://` links pointed at the default gateway on web.** The shared renderer asks the platform for a gateway host; web supplied no hook, so a self-hosted site rendered `hyper.media` hrefs. Only the href was affected — clicks are intercepted either way — so it showed up in hover, copy-link, and open-in-new-tab.
- **A duplicated `vi.mock`**, where two distinct specifiers collapsed onto one path during the move.

## Follow-up, not in this PR

23 desktop test files now exercise shared `@shm/ui/agents` code while `@shm/ui` has no agents tests of its own, so the shared package's coverage only runs with the desktop suite. The pure-logic ones (session rows, chat parts, model utils, markdown, run-work) could move to the package; the desktop-integration ones should stay.

`TERMINAL_RUN_STATUSES` is defined in both `run-work.tsx` and `rich-message-composer.tsx`. That duplication is on main already and is carried forward unchanged rather than fixed here.


ericvicenti:
Rebased onto current main (c1fa83fa9) — not mechanical, since 19 commits landed on main touching the very files this PR moves into `@shm/ui/agents` (collaborators tab #959, scheduled work, run-tree rendering, hover-reveal avatars, tools management). Rename-aware merge carried main's changes into the relocated files; conflicts were resolved keeping main's newer behavior with this branch's paths.

Three follow-up commits from the rebase:

- **fix(ui): shed imports main left dead** — main carries a few unused imports that desktop's tsconfig tolerates but `@shm/ui`'s flags, plus one `QueuedChatMessages` import whose last caller main removed in #959. Also one `noUncheckedIndexedAccess` fix.
- **fix(web): parse every agent tab the route schema allows** — main's new `collaborators` tab was missing from the web URL parser's hardcoded list, so a shared `?tab=collaborators` link silently landed on the default tab. The list now derives from the route schema, with a round-trip test over every tab.
- **fix(desktop): avatar test asserts the label actually rendered** — pre-existing red on main: #959 wrote the test, then the clickable-avatars commit changed the aria-label without updating it.

Verification: workspace typecheck clean, format clean; desktop 688/688, `@shm/ui` 314/314, web 161/162 (1 pre-existing skip), shared suite green; web production build passes with the agents chunk still code-split (318 kB, only on agents routes). Also smoke-tested live: `./dev run-web` + a local agent server on 3051, driven headless — signed-out gate renders, sign-in dialog opens, and signed in the list shows the green server health dot and Create Agent talks to the live server (correctly asking for a model provider first). Zero console errors on the agents routes.

ericvicenti:
Two more pushes since the rebase:

**The agents-list gear works on web** (`5daa0922c`). It navigated to desktop's Settings window — a route web can't open, hence "failed to open route". `AgentServersSettings` moved into `@shm/ui/agents/server-settings` (it was already platform-neutral), desktop's gear keeps opening its Settings window via a new optional seam member, and web opens the same management UI in a dialog right on the agents page. Verified live: add/remove/make-default all work in the browser.

**Web signs as the vault account it was delegated — same agents on both surfaces** (`df80c5ccb`). Desktop signs with the account key; web signed with its device key *and reported the device key as the account*, so the same person had two disjoint agent lists. The server already accepted delegated signers from its `account_authorizations` table — nothing populated it. A new `RegisterSigner` action carries the vault-issued Capability blob; the server verifies the blob's signature (account granted it), that the delegate is the envelope signer (caller holds the key), and the role. Web now reports the delegated account, and the shared client self-registers on first rejection and retries. Proven end to end: an agent created over the wire by a raw account key appears in a headless browser holding only the delegated device key.

Two operational notes:
- **Deploy the agents server before (or with) the web app.** A delegated web client against an old server gets "Unsupported action: RegisterSigner" instead of its agent list. The agentic box auto-deploys on image push, so in practice: merge, let the server image roll, then deploy web.
- **Agents previously created *from web* live under the device-key identity** and won't automatically migrate under the account. Desktop-created agents (the common case) just appear on web once this lands.

Suites: agents 321/321 (5 new auth tests incl. capability replay/tamper rejection), desktop 688/688, ui 314/314, web 165 (4 new adapter tests), typecheck/format/build clean.

PR #980, opened 2026-08-25
Slot block for root level mixed grouping

iskaktoltay:
Adds a new **Slot** block type: an invisible grouping wrapper, so lists and blockquotes can live at the document root, where there's no parent block to carry a childrenType. This enables mixed grouping at the top level.

### What's included (by commit)
- Slot block type and registration: the node, schema, and invisible rendering.
- Paste, merging and boundary normalization: a SlotNormalization ProseMirror plugin that repairs orphaned/malformed slots, merges adjacent same-grouping slots, unwraps non-root slots. Paste handling so pasted lists land in a valid Slot.
- Markdown export: root Slot lists serialize as normal markdown lists.
- Slot-aware multi-item unnest: unnest is now fully slot-aware and consolidated onto one path (unnestBlock).
- listLevel removal: unordered markers now cycle from real nesting depth via CSS.
- Testing: updateGroup, SlotNormalization, normalizeFragment, block-group-map, and updated nestBlock/unnestBlock/splitBlock/nodeConversions fixtures.

### DIFF
--- a/frontend/packages/client/src/blocks-to-markdown-slot.test.ts
+++ b/frontend/packages/client/src/blocks-to-markdown-slot.test.ts
@@ -1,0 +1,39 @@
+import {describe, expect, it} from 'vitest'
+import {blocksToMarkdown} from './blocks-to-markdown'
+
+// A root-level Slot wrapping a list of paragraph items.
+function slotDoc(childrenType: string, items: string[]) {
+  return {
+    metadata: {},
+    content: [
+      {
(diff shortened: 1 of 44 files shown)


horacioh (approved):
LGTM. please fix the failed tests. but I tried the editor and it does not feel that wierd. we might have some issues with adding stuff inside headings and so on, but nothing too dramatic!

great work @iskaktoltay !

PR #983, opened 2026-08-25
Rename the "site" concept to "space"

ericvicenti:
Product language settled on **space** for a published account home, but the code still said *site* nearly everywhere. This renames the concept across the Go backend, desktop, web, notify, vault, CLI, agents and the shared/UI packages — identifiers, components, hooks, file names, comments and user-facing copy.

348 files, ~3.7k lines. Most of it is mechanical; the parts worth reviewing are the compatibility boundaries below.

## What deliberately still says "site"

Anything that crosses a version boundary keeps the old spelling, commented at each site:

| Surface | Kept | Why |
| --- | --- | --- |
| Contact blob | `subscribe.site` (`refmt:"site,omitempty"`) | signed CBOR |
| Document metadata | `siteUrl` | persisted key, also a SQL attribute lookup |
| Notify service | `siteUid`, `mark-site-read`, `notifySiteDiscussions`, `site-doc-update`, `site-new-discussion` | signed payloads, sqlite columns, stored notifications |
| Seed Host API | `sites` endpoint, `siteCount`, `currentSiteUrl`, `siteUid` | external service |
| API payload | `hasSite` | `.strict()` schema — see below |
| Agent triggers | `site-update` | see below |
| Ops | `SEED_SITE_*`, `seedhypermedia/site`, `seed-site`, `~/.seed-site`, `sites.csv`, the monitord `-sites` flag, the `source="site"` Prometheus label | `ops/` is untouched entirely |
| Web platform | `sec-fetch-site`, `same-site`/`cross-site`, `site-per-process` | not ours to rename |
| English | "call site", "Import Web Site", MediaWiki `meta=siteinfo` | different word |

## Renames that came with an upgrade path

- **Desktop route keys.** `site-settings` → `space-settings` etc. Routes are persisted in `WindowState-v004`, and `openInitialWindows` drops any window whose routes fail `navRouteSchema` — so a stale key would have cost the user the *whole window*, not just a blank page. `migrateLegacyRoute` upgrades the keys (plus `explore.context.type` and `library.grouping`) before that validation runs. Covered by four tests.
- **HTTP routes.** `/hm/create-site` 301s to `/hm/create-space`; `/hm/api/site-image` aliases the new action so a browser on the previous bundle can finish an upload mid-deploy. Both stay in `RESERVED_HM_SUBROUTES`.
- **Auth params.** `space_name` / `spaceName` are written *alongside* the old keys and read either way, so a vault on an older build still shows the space name.
- **CLI.** `--site-url` stays (it maps straight to the `siteUrl` metadata key); `--space-url` is added beside it.
- **Published client.** `@seed-hypermedia/client` ships deprecated aliases for `PARAM_SITE_NAME`, `HMSiteMember(Schema)`, `siteDiscoverRequestSchema`, and still accepts `siteName` on its auth config.

## Two follow-ups this PR does not take

**`hasSite`** — `HMMetadataPayloadSchema` is `.strict()`, so an unknown key is a hard parse failure in *both* directions: a new client would reject `hasSite` from an older server, and an already-published CLI would reject `hasSpace` from an upgraded one. Emitting both doesn't help — the old strict schema rejects the extra key. Flipping it needs a release that accepts either, then a second that switches.

**`site-update` agent triggers** — I renamed this, then reverted it during review. The agents service deploys independently of the desktop app, and desktop builds predating this PR do `TRIGGER_TYPE_ICONS[context.source.type]` then render `<Icon />` with no fallback — an unknown type is `undefined`, which throws *Element type is invalid* and takes out the triggered-session card. Same two-phase fix applies: ship a desktop release that accepts both spellings, then flip the server.

## Verification

- `pnpm typecheck` clean across all 16 projects; `vault` clean
- `pnpm format:check` clean
- `go build ./...` and `go test ./backend/...` clean
- Tests: web 197, shared 1032, desktop 692, ui 325, notify 94, agents 134
- `pnpm --filter @shm/web build` succeeds with all four routes in the manifest

Pre-existing failures, unrelated and untouched by this branch: `client.test.ts` (Zod `nextPageToken`), `client/vault-local.test.ts`, `editor/readonly-viewer-gallery.test.tsx`, and the `agents` typecheck error at `api-service.test.ts:5203`.

https://claude.ai/code/session_01VpayUJJV7RDovKerpaWftB


ericvicenti:
Closing in favor of a narrower change: renaming site → space in user-facing copy only, with no code, wire, or identifier changes. The full rename is more churn than it's worth right now, and this PR's size also broke lint-go's only-new-issues filtering (GitHub caps diffs at 300 files).

PR #987, opened 2026-08-26
Web: agents assistant panel beside every site page

ericvicenti:
## Summary

Brings the desktop assistant sidebar to the web app, sharing the component itself.

### Shared (`@shm/ui/agents`)
- `assistant-panel.tsx`, `assistant-selection.ts`, `assistant-session-ref.ts`, `assistant-window-context.ts` moved from desktop into the shared package (via `git mv`). The panel now uses the platform seam (`./account`, `./navigation`, `AgentsRichEditorSubmitHandle`) and gained an optional `onClose` close button. Desktop imports it from `@shm/ui/agents/assistant-panel`; behavior unchanged, all desktop assistant tests pass (116).
- **Latent web bug fixed:** `models.ts` wrote optimistic sessions/messages to the `queryClient` module singleton, but web renders with its own registered client, so those writes were invisible on web. Added `getQueryClient()` and routed all cache writes through it.

### Web
- **"Agents"** item in the account avatar menu (dropdown + mobile sheet copies) toggles the panel; reads "Close Agents" while open.
- `web-assistant-host.tsx` — hosted inside `WebSiteProvider` beside every page: the page column always renders (toggling never remounts the page), plus a draggable divider (280 px min, 40 % of viewport max, width persisted) and a sticky full-height aside. Narrow screens get the shared `MobilePanelSheet`.
- `assistant-panel-state.tsx` — root-level context (`isOpen`, `sessionId`, `newChatRequest`) persisted to localStorage, the web analogue of desktop's window state.
- `web-assistant-panel-content.tsx` — client-only lazy chunk: registers the platform, sign-in gate, shared panel.
- Divider CSS fixed (hover color referenced an undefined `--color8`).

### Bundle
Production build: the 458 KB agents chunk is only statically imported by the two lazy chunks (`agents-page-content`, `web-assistant-panel-content`). The always-loaded `providers` chunk is 18 KB with no agents code — just a dynamic `import()`.

## Test plan
- [x] ui / desktop / web typechecks
- [x] desktop unit tests (assistant suites 116/116), web unit tests (36 files, 203 tests; 5 new for the host/state under StrictMode)
- [x] Driven in a real browser against the dev stack: signed-out panel, divider drag + persistence, close; signed-in avatar menu → Agents → panel with agent picker → Close Agents; mobile bottom sheet. No console errors.
- [ ] CI

https://claude.ai/code/session_01K3qMAzkS2n5WXESGxVy1Wi


ericvicenti:
Follow-up commit: the panel is now hosted above the Remix outlet (no remount/flash on page transitions — verified the aside is the same DOM node across navigations) with a small site-context bridge so the window-context feature keeps tracking the current page; on narrow screens it's full-screen with a "Back to page" bar instead of a bottom sheet.

PR #1000, opened 2026-08-27
feat(documents): collections are a shape the indexer derives, not a metadata flag

ericvicenti:
## What

A document is a **collection** when its content is a single childless `Query` block listing that document's own children. Until now the only thing that said so was `"type": "Collection"` metadata (#994), which means:

- every collection authored before that flag existed is invisible as one — e.g. [`hm://z6MkiAK…/notes`](https://explore.hyper.media/hm/z6MkiAKDcRSzQ4zPZfnJcS5HYx5MwgN6MU9foHihJGrhqNBj/notes), which has the shape and no flag
- a stray metadata key could claim to be a collection while looking nothing like one

Shape is now the single source of truth, derived by the indexer and exposed as `DocumentInfo.is_collection`.

## Why the indexer

Resolving a document's block tree means replaying its whole change history — `docmodel` territory, and `blob` can't import it. So the deriver is injected, the way the fallback-cover deriver already is.

Crucially it **rides that existing replay** rather than adding a second one. `DeriveFirstContentImage` becomes `DeriveDocFields` returning `DerivedDocFields{FirstImage, IsCollection}`: one model rebuild, both fields. The rebuild is the entire cost; each field is then a walk over a tree that already exists.

One consequence worth knowing: the replay is no longer gated on cover/icon (that gate made sense when the cover was the only derived field — a collection may well have an icon). Storing `FirstImage` is still gated.

## Why an async backfill and not a reindex

A full reindex truncates every derived table and rebuilds from every blob **in one write transaction**, holding the only write connection and blocking startup in `State_MIGRATING`. That's the right hammer for a field the read path can't function without. It's the wrong one here:

- an underived flag reads as *"not known to be a collection"*, which degrades gracefully
- the write path derives for every new and updated document regardless

Only history needs filling in, and it can be filled in late without anyone noticing. So `backend/blob/backfill.go` runs like the embedding indexer: bounded batches (50), one transaction each, 2s between passes, started after migrations and detached from the run group. Presence of the `$db.isCollection` attribute row **is** the bookkeeping — anti-joined on `document_attributes`' `(resource, key)` primary key — so there's no progress table to keep in sync.

Both #908 failure modes are contained:
- a document that fails to derive records `false` rather than staying absent, so the backfill never re-picks the same undecodable blob forever. The cover sentinel is deliberately *not* stamped there: empty means "derived, no image" and a failed read gives no grounds to claim it.
- five consecutive infrastructure failures stop the worker instead of warn-looping for the process lifetime.

## Client side

`document-machine` keeps its own copy of the predicate (`isCollectionContent`) because a draft has never been indexed. It matches the Go version: any self-targeting include, either mode (`AllDescendants` includes the direct children too) — same semantics as the existing `documentHasSelfQuery`.

"Convert to document" can no longer clear a flag, so it prepends an empty paragraph. The query block survives and renders inline — exactly where clearing the metadata used to leave it.

## A bug found on the way

Listings rebuild their attributes from a SQL json aggregate, and SQLite has no boolean type, so a kind-`"b"` attribute arrives as the **number** `1`/`0` and a plain `.(bool)` assertion silently fails. `indexedAttrBool` reads both representations.

## Testing

- 14 docmodel unit tests: both path encodings (`notes` vs `/notes` — both exist in the wild), both modes, multi-include, root documents, unretargeted draft placeholders, and each disqualifying shape
- `TestListDirectoryDerivesIsCollection`: end to end through `ListDirectory`, including that adding a second top-level block clears a stale flag, and that `"type": "Collection"` metadata no longer overrides shape
- `TestBackfillDerivesIsCollection`: deletes the attribute to simulate a pre-upgrade database, then asserts unset ≠ false, that a bounded pass respects its batch size, and that a drained pass reports zero
- Green: `backend/blob`, `backend/api/documents/...`, `backend/storage`, `@shm/shared` (1034), `@shm/ui` (356), `@shm/desktop` (716), plus `tsc` on client/shared/ui/desktop

## Reviewer notes

**Codegen is verified against the real toolchain.** `./dev gen` is a no-op on this branch, and both `//proto/documents/v3alpha:go.check` and `:js.check` pass their actual hash comparison (forced past the mtime shortcut, with a corrupted-gensum negative control to confirm the check really fires). Backend tests were re-run under the repo's pinned Go 1.26.2, including `backend/daemon`, which exercises the boot path the backfill worker is wired into. `pnpm format:check` passes across every workspace. The integration suite (`./dev test:integration`) has not been run.

**One thing to decide.** `normalizeCollectionEditorBlocks` currently *appends* a query block to whatever blocks exist, so a #994-era collection can legitimately have prose plus a query block. Under a strict single-block predicate those stop reading as collections. Nothing in the tests covers that case and I don't know whether any exist in practice, but if they do, the predicate needs loosening to "exactly one *self-querying* top-level block" or similar — worth a look before this ships.

https://claude.ai/code/session_01YZYw3xXjs8mfd7rb3M97Sj


horacioh:
I renamed from collection to folder. I hope that's ok

ericvicenti:
@horacioh will take this over, probably in #998

horacioh:
yes. thanks for the help @ericvicenti !

PR #1016, opened 2026-08-30
feat(agents): ~/agents directory for delegation-target discovery

ericvicenti:
## What

`read ~/agents` — a new introspection address listing the account's agents: id, name, active model (`provider/model`), and enabled models per agent. This is exactly the set `delegate {agentId}` may target, closing the discovery gap: agents could already run children under sibling agents, but had no way to learn their ids, so cross-agent delegation only worked when the user pasted an id into the chat.

## Privacy posture

**Public agents are fenced off from private siblings.** A public agent's transcripts are readable by every signed account, and `public_chat` lets outside accounts drive its sessions, so:

- `read ~/agents` from a public agent lists **only public siblings** — the listing lands in a world-readable transcript and must never name private agents.
- `delegate {agentId}` from a public parent to a private sibling is **refused** (`requireDelegationVisibility`), enforced at both spawn sites (chat delegation and workflow `ctx.delegate`) — a public transcript would otherwise carry the private child's brief, title, and result, and an injected chat message could steer a more-privileged private agent.
- Private agents still see and may delegate to the whole account, public siblings included.

The directory is also summaries-only regardless of visibility:

- Only agents the account **owns** are listed — the same gate `#spawnSubSession` enforces on `agentId` (shared/collaborator agents are not delegatable, so they are not listed).
- A sibling's system prompt, grants, tool set, and signing-key names are **not** exposed; tests assert prompt text and key names never appear in the payload.
- Guidance in the payload explains what delegation to a listed agent means: the child runs under that agent's own prompt, tools, model, and memory.

Documented in `security.md` (public/private boundary) and `tools.md` (address table).

## Notes

- Complements #1015 (delegate `model` parameter): the directory surfaces each sibling's `enabledModels`, so once both land, an agent can route a task to a sibling *and* pick an appropriate tier from that sibling's enabled set. No code dependency, but both PRs touch `#spawnSubSession`, so whichever merges second needs a trivial conflict resolution.
- Tests in `verbs.test.ts` cover the directory listing, the summaries-only guarantee, the public filter, and the delegation boundary in all directions; full agents suite 390 pass, typecheck clean.

https://claude.ai/code/session_013bwThptdnZqMbQMzd1h5jF


ericvicenti:
Closing: we've decided against direct agent-to-agent communication for now — agents will collaborate through Seed content (docs/comments) instead, where the capability system already governs access. Cross-agent delegation (delegate {agentId}) is being removed in a follow-up PR, which makes this directory moot.

PR #1019, opened 2026-08-30
fix(daemon): serve IPFS files with a Content-Type derived from their bytes

ericvicenti:
## Problem

`GET /ipfs/{cid}` answered every request with `application/octet-stream`. Chromium sniffs raster images inside ``, so JPEG/PNG worked, but it deliberately does not sniff SVG — so every SVG image (covers, icons, inline images) rendered blank in the desktop app. Example: https://hyper.media/hm/z6MkpVa5nMUR5ZaUEyV1SE48KTNbwTuHRd83RwLgMKc4nGU3 — 18 of the 22 image placements on that page are two SVG blobs. Web was unaffected only because `/hm/api/image` rasterises everything through sharp.

The same missing header also caused:
- **Export as Markdown** naming every media file `foo.bin` (`mime.getExtension('application/octet-stream')` → `bin`)
- **Save file** dropping the extension when the block name has none (it already special-cases octet-stream)
- the **inspect page** showing SVGs as text (its "is this an image?" probe is an `` load)

Not a sync issue — all CIDs were already local on both daemons.

## Fix

Detect the type from the leading bytes at the one place that serves them, using `gabriel-vasile/mimetype` (already an indirect dependency). This deliberately does *not* add a filename or mime attribute to blocks: that would be hidden data, redundant with the content, and something every writer (desktop, CLI, agents, mobile, importers) would have to keep correct forever. The type is a property of the bytes.

Precedence in `ContentTypeFor`:
1. `?filename=` extension (download name — visible to the user, already used by the File block)
2. byte detection
3. `application/octet-stream`

## Safety

Serving real types from the daemon origin (`localhost:56001`, which also hosts the API) means an uploaded SVG or HTML with `<script>` could otherwise execute against the daemon's own API when navigated to directly. So:
- `X-Content-Type-Options: nosniff` on every `/ipfs/` response
- `Content-Security-Policy: sandbox` on anything that isn't a raster image, video, audio, font, or PDF (SVG included — it can carry script)
- `text/html` / `xhtml` downgraded to `text/plain`

`` loads are unaffected: images never execute script. The web `/hm/api/file` proxy forwards the two headers so the same guarantee holds on the site origin.

## Tests

`backend/hmnet/filemanager_contenttype_test.go`: precedence/downgrade/inert-set unit tests, plus handler tests for SVG (typed + sandboxed), PNG (no sandbox), HTML → plain text, unknown → octet-stream, `?filename=report.pdf`, and a Range request on an SVG keeping type and offsets. Existing `TestPostGet`/`TestRangeRequests` still pass.

- Web's `/hm/api/image` could stop rasterising SVGs now that the daemon reports `image/svg+xml`.
- `save-cid-as-file.tsx`'s octet-stream special case can go once this ships.

https://claude.ai/code/session_011fyabXL7WrNoKcTjioRuSc

### DIFF
--- a/backend/hmnet/filemanager.go
+++ b/backend/hmnet/filemanager.go
@@ -6,13 +6,16 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"mime"
 	"net/http"
+	"path"
 	"seed/backend/blob"
 	"seed/backend/ipfs"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/gabriel-vasile/mimetype"
 	"github.com/ipfs/boxo/blockservice"
 	blockstore "github.com/ipfs/boxo/blockstore"
 	"github.com/ipfs/boxo/exchange"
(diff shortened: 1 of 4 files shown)


ericvicenti:
@burdiyan I hope you don't mind that we do this

The problem is that SVG images are not being rendered correctly in the desktop app because the mime type is not being set by the server. The server has no clue what mime type to send because it is opaque binary data, so my agent suggested this library which is able to infer many mime types correctly.

burdiyan:
I thought we were already inferring mime-types. I remember having done something about some time ago. Hopefully this doesn't do it on every request, because sometimes we do know the expected file name or the extension when we are loading the image from a document, right? If we load them with `filename` query param it then should construct the mime type more efficiently perhaps, without having to sniff the bytes.

But otherwise I've always wanted to have a fallback to detect mime type by looking at the bytes.

PR #1024, opened 2026-08-31
fix(agents): keep the API responsive under concurrent runs

ericvicenti:
## What happened

Prod agents server became unusable today (30–45s for a bare 404) while Ion ran a 9-way parallel audit workflow. Diagnosed live on the box:

1. **Memory-summary walks on every turn** — Ion's memory held **201,492 files** (an imported `src/` tree, `toolchains/`, `.pnpm-store`, …). `buildSpaceIndex` → `summarizeMemoryTopLevel` walks up to 10k entries synchronously (~175ms measured), and every memory write invalidates the cache — 9 concurrent runs kept it walking continuously (60k `statx`/10s observed via strace).
2. **SQLite page-cache thrash** — after quarantining the giant memory dirs, strace showed **169k `pread64` in 10s**, all on `agents.sqlite` (288MB vs bun:sqlite's default 2MB page cache). All synchronous, all on the only thread.
3. **Thread search full-scans** — `ORDER BY e.created_at DESC LIMIT 4000` had no index: a temp b-tree sort over every event blob (131MB) per call.

With 8 concurrent model runs + microsandbox VMs pinning 4 host cores, loop utilization hit ~100% and request latency exploded by queueing.

## Fixes

- **Cost-proportional hold on memory-summary walks**: a walk earns a hold of 100× its duration (capped 60s; <5ms walks stay exactly fresh). A pathological memory can no longer monopolize the loop.
- **SQLite pragmas**: 64MB page cache, 512MB mmap, in-memory temp store.
- **New migration + baseline index** `session_events_by_created (created_at DESC)`; EXPLAIN confirms the thread-search now walks recency order and stops at its LIMIT.
- **Configurable run-queue caps**: `SEED_AGENTS_MAX_CONCURRENT_MODEL_RUNS` / `SEED_AGENTS_MAX_CONCURRENT_WORKFLOWS` (defaults unchanged: 8/32). Recommend setting model runs to ~3 on the current 4-core prod box.

## Ops already applied on prod (reversible)

- Moved Ion's `memory/{src,toolchains,worktrees,.pnpm-store,.cache}` (→ 201k→352 files) to `/data/agents/<agent>/memory-quarantine/` with a `QUARANTINE.md` note.
- Restarted `agents-stable` with the queued audit runs deferred (`not_before` = 20:28 UTC). Nothing cancelled.
- Latency recovered: 30–45s → ~1s.

## Testing

- `bun test`: 401 pass / 0 fail; `bun tsgo --noEmit` clean.
- Query plan verified on a migrated in-memory DB.

https://claude.ai/code/session_014fkiUq95srysiati9AsjzU

### DIFF
--- a/agents/src/agent-memory.test.ts
+++ b/agents/src/agent-memory.test.ts
@@ -12,6 +12,7 @@ import {
   readMemoryFile,
   resolveMemoryPath,
   summarizeMemoryTopLevel,
+  summarizeMemoryTopLevelAsync,
   writeMemoryFile,
 } from '@/agent-memory'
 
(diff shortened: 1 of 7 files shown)


ericvicenti:
Second commit, per Eric's direction: the memory rollup is now **shallow (2 levels) and off the prompt path**.

- `summarizeMemoryTopLevel` walks at most 2 directory levels (`MAX_MEMORY_SUMMARY_DEPTH`) — the prompt line only names root entries with rough counts, so deep recursion into a pnpm store or imported source tree bought nothing. Deeper trees read as minimums (`truncated`). The depth budget skips a subtree but keeps walking siblings; only the entry budget aborts.
- New `listMemoryAsync`/`summarizeMemoryTopLevelAsync`: every readdir/stat awaits, so a big walk shares the event loop instead of blocking it.
- `buildSpaceIndex` is now adaptive: a memory whose last walk cost >5ms never walks synchronously again — it serves the last line and refreshes in the background (one in-flight refresh per state dir; cached space indexes are dropped when the line changes). Small memories still walk sync and stay exactly fresh.
- The exec before/after snapshots and the Memory tab keep their full-depth (capped) walks — the exec diff needs deep paths to report changed files.

402 tests pass, typecheck clean.

PR #1025, opened 2026-08-31
Make child and trigger prompts explicit

ion-lion:
## Real-world failure

Every model-backed trigger currently starts with the agent's ordinary interactive system prompt and full tool grants. The trigger can only customize its first *user* message. That makes a scheduled auditor, comment responder, or maintenance worker behave like the general chat agent even when it needs a deliberately narrow role and tool posture.

Child delegation technically had a system-prompt override, but exposed it as the overloaded `prompt` field next to the actual task `brief`. We already carry compatibility logic because models sometimes put the brief in `prompt`; the API therefore makes one of the most important execution controls ambiguous.

This is reachable today in every `newThread` trigger and every model child.

## Root cause

Prompt/tool overrides existed only in the private child `SubSessionSpec`. Trigger firings enqueued a normal session-message run with no run-specific definition profile. The public delegate schema also named its system override `prompt`, while trigger `prompt` means the first user message.

## What changes

- Adds explicit `delegate {systemPrompt}` while retaining `prompt` as a compatibility alias. Specialized instructions append to the ordinary agent prompt by default; `includeAgentSystemPrompt: false` opts out of that authored prompt.
- Lets `continuation: {kind: "newThread"}` carry:
  - `systemPrompt`: additional system instructions for that firing.
  - `includeAgentSystemPrompt`: whether to include the ordinary agent-authored prompt (default `true`).
  - `tools`: an allowlist that is intersected with the agent's grants, so it can narrow but never escalate privileges.
- Stores the trigger profile in the durable run input and applies it through the same definition-narrowing path used by delegated children.
- Preserves already-persisted child runs that still contain the old `spec.prompt` field.
- Updates the trigger write guide and delegate tool contract.

The host's shared Seed/runtime instructions still append normally. With the default `includeAgentSystemPrompt: true`, child/trigger instructions follow the ordinary agent-authored prompt; setting it to `false` omits only that authored prompt.

## Before / after

Before:

```json
{
  "prompt": "Audit the latest changes",
  "continuation": {"kind": "newThread"}
}
```

The firing uses the normal chat system prompt and grants.

After:

```json
{
  "prompt": "Audit the latest changes",
  "continuation": {
    "kind": "newThread",
    "systemPrompt": "You are a read-only release auditor. Report evidence, not guesses.",
    "tools": ["read", "search"]
  }
}
```

The firing gets the specialized system role and a narrowed callable set. Children use the same explicit vocabulary via `delegate {brief, systemPrompt, tools}`.

## Validation

- Focused API tests after rebasing onto current main: **9 passed, 0 failed**.
  - Trigger profile survives create/read normalization.
  - Trigger and child prompts include the ordinary agent prompt by default, while the explicit opt-out omits it at the provider boundary.
  - Trigger tool allowlist is stored in the durable run input.
  - Explicit child `systemPrompt`, legacy `prompt`, and lone-`prompt` brief compatibility are covered.
- `bun run format:check`: passed.
- Full `api-service.test.ts`: **121 passed**; three unrelated environment-sensitive tests failed (one pre-existing parked-child 5s timeout and two local MCP discovery tests whose loopback server was unavailable in the isolated executor).
- Full `tsgo --noEmit` could not complete in the 300-second / constrained-memory execute sandbox; a production-only attempt was SIGKILLed. Focused tests compile and execute the changed paths; CI should run the canonical check on normal runners.

## Limits / follow-up

- This config specializes model-backed `newThread` firings. Headless tool/script continuations already wire tools directly and invoke models explicitly through `ctx.delegate`.
- This PR does not add trigger-specific model selection or a visual form for the advanced fields; the authored trigger API/write tool is the first surface.

## Provenance

[Originating agent session](https://seed.hyper.media/hm/agents/session/eb90a676-31fd-44a3-9709-65b369a65b0d?server=https%3A%2F%2Fagentic.seed.hyper.media&agent=0d941d24-aca1-4eb9-990c-0a41097e9667)


ericvicenti (changes requested):
i think there should be a toggle on the trigger and child prompt to include the system prompt or not.

by default it does include system prompt, and the trigger/child session will include it

ion-lion:
Addressed in `ba44cba69`.

- Added `includeAgentSystemPrompt` to both model children and `newThread` trigger profiles; it defaults to `true`.
- A child/trigger `systemPrompt` now appends specialized instructions after the ordinary agent-authored prompt by default.
- `includeAgentSystemPrompt: false` omits only the agent-authored prompt; shared Seed/runtime instructions remain.
- The trigger toggle is normalized, persisted in durable run input, and applied through the same provider prompt path as children.

Validation: focused API prompt tests **9 passed**, including provider-boundary assertions for default inclusion and explicit opt-out; protocol tests **3 passed**; changed-file Prettier and `git diff --check` pass. Full `tsgo --noEmit` again exceeded the 300-second executor limit.

Issue #901, opened 2026-07-28
CLI: document move reports success when signer has no capability (write silently not materialized)


ericvicenti:
## Summary

`seed-cli document move` prints `✓ Document moved: <url>` even when the signing key has **no WRITER capability** on the target space. The server accepts the published ref blobs without an authorization error, but never materializes them — so the user is told the move succeeded when nothing actually changed.

## Repro

Keyring contains a single key that is **not** a delegate on the target space:

```
$ seed-cli key list
[ { "name": "main", "accountId": "z6MkmZUb4K5c17zGGBuJJerwFzBaGkiYLfEEnkb9CH1W1ptb" } ]
```

Move a doc in a space where that key has no capability (verified via `account capabilities` — 10 WRITER delegates, none of them this key):

```
$ seed-cli -s https://seedteamtalks.hyper.media document move \
    "hm://z6MkuBbs…/tech-talks/customizations-with-code" \
    "hm://z6MkuBbs…/tech/customizations-with-code"
✓ Document moved: https://seedteamtalks.hyper.media/hm/z6MkuBbs…/tech/customizations-with-code
```

But:

- `document get` on the destination → `not-found` (metadata `null`)
- `document get` on the source → still returns the original document, no redirect

## Cause

In `frontend/apps/cli/src/commands/document.ts` (the `move` action, ~line 704–760), the command:

1. `client.publish(versionRefInput)` — destination version ref
2. `client.publish(redirectRefInput)` — source redirect ref
3. `printSuccess('Document moved: …')`

Both `publish` calls resolve without error because the server accepts the signed blobs regardless of whether the signer is authorized to write to the space; the refs are simply never materialized. The CLI treats "blob accepted" as "operation applied."

This is not specific to `move` — any write command that goes through `client.publish` (`create`, `update`, `fork`, `redirect`, `delete`, comment create, …) presumably has the same silent-failure mode with an unauthorized key.

## Expected

Either (ideally both):

1. **Server-side:** the publish endpoint should reject ref blobs whose signer lacks capability on the target space with a clear authorization error, so clients can surface it.
2. **CLI-side:** after publishing, verify the operation materialized (e.g. fetch the destination and check the version, or pre-check the signer against `account capabilities` for the space) before printing success, and exit non-zero with a "signer has no WRITER capability on this space" error otherwise.

## Environment

- `@seed-hypermedia/cli` via `npx -y @seed-hypermedia/cli` (latest published, 2026-07-28)
[...]

PR #1055, opened 2026-09-06
fix(cli): authorize both sides of document moves

ion-lion:
## Why

`seed-cli document move` could publish both move refs without attaching delegated capabilities. Because the daemon can accept blobs before asynchronously rejecting unauthorized refs, the command could print `Document moved` even though a cross-account move was not authorized. In the worst ordering, the destination ref could be accepted while the source redirect was rejected, leaving a partial operation. This addresses #901.

## What changed

- Resolve authorization for both source and destination paths before publishing either side of a move.
- Fail immediately when the selected key owns neither account and has no WRITER/AGENT capability.
- Attach the destination capability to version or republish refs.
- Attach the source capability to the redirect ref.
- Add an integration regression proving an unauthorized foreign source is rejected, no success is printed, the source remains readable, and the destination does not materialize.

## Behavior

Before: cross-account move refs omitted capabilities and the CLI could report success for an unauthorized or partial move.

After: predictable missing authorization fails before the first publish; same-account moves retain their existing behavior, while delegated moves carry the capability CID used for each affected account/path.

## Validation

- `git diff --check` passed.
- Independent exact-diff review traced `resolveCapability`, `createVersionRef`, and `createRedirectRef` types and found no blocker across same-account, delegated source/destination, and republish flows.
- The pinned Node/pnpm/Bun toolchain was installed, but the monorepo's filtered pnpm install did not complete within repeated 300-second sandbox limits, so local `tsc` and the daemon integration fixture could not run here. CI is required for executable validation; this PR remains draft until it passes.

## Limits / follow-up

This prevents the known missing-capability failure before publication. It does not make the two-publish move atomic against process failure or a later indexing failure; a broader materialization/transaction design should be handled separately rather than hidden in this fix.

### Current validation (2026-09-10)

- Rebased conflict-free onto current upstream main `1d35e974b8276020133a5385d6f49ff776b58598`; exact branch head is `f4f740960782f6f20402b7e37f65004d46828b26` and matches `ion-lion/seed`.
- `git diff --check origin/main...HEAD` passes; the patch remains limited to the move command and its daemon-backed regression fixture.
- A fresh current-main code trace confirms the issue is still reachable: daemon ingestion intentionally stashes unauthorized refs for out-of-order P2P authorization, so the CLI must resolve both capabilities before publishing rather than treating blob acceptance as materialization.
- Exact-head GitHub Typecheck, Lint, shared/desktop/web unit tests, editor E2E, and aggregate check all pass. The generic `integration-tests` job was skipped. The daemon-backed CLI regression remains covered by the companion CI proposal in #1057 and is not claimed as executed on this head.

### Validation status (2026-09-06)

- Exact head `cec7024947f25eb0abcfe5248ef03ec08732c667`: required Typecheck, Lint, shared/desktop/web unit checks, editor E2E, and aggregate check passed.
- The workflow's `integration-tests` job was skipped (`if: inputs.run-integration-tests` on the direct PR run), and that job runs `tests/`, not `frontend/apps/cli/src/test/cli-fixture.test.ts`; therefore the added CLI daemon fixture did **not** run in CI.
- I attempted the exact Bun test locally. Test loading was repaired and the fixture reached daemon startup, but the constrained runner could not compile the required llama.cpp native library within its 300-second execution ceiling, so the daemon never became ready. This is an environment/setup failure before the test body, not a passing claim.
- `git diff --check` passes and the branch is clean.

### DIFF
--- a/frontend/apps/cli/src/commands/document.ts
+++ b/frontend/apps/cli/src/commands/document.ts
@@ -811,6 +811,27 @@ export function registerDocumentCommands(program: Command) {
         const key = await resolveSigningKey(_options.key, keyOptions(globalOpts))
         const signer = createSignerFromKey(key)
 
+        // Resolve both authorizations before publishing either side of the move. The daemon can
+        // accept a Ref blob before asynchronously rejecting it, so publishing first can otherwise
+        // leave a destination behind and let the CLI report success even though the source did not
+        // move.
+        const sourcePath = hmIdPathToEntityQueryPath(source.path)
+        const destPath = hmIdPathToEntityQueryPath(dest.path)
+        const sourceCapability = await resolveCapability(client, source.uid, key.accountId, sourcePath)
+        if (source.uid !== key.accountId && !sourceCapability) {
+          throw new Error(
+            `No WRITER or AGENT capability found for key ${key.accountId} on source account ${source.uid}. ` +
+              `Use "account capabilities hm://${source.uid}" to check available capabilities.`,
+          )
+        }
+        const destCapability = await resolveCapability(client, dest.uid, key.accountId, destPath)
+        if (dest.uid !== key.accountId && !destCapability) {
+          throw new Error(
+            `No WRITER or AGENT capability found for key ${key.accountId} on destination account ${dest.uid}. ` +
+              `Use "account capabilities hm://${dest.uid}" to check available capabilities.`,
+          )
+        }
+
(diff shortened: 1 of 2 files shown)


ion-lion:
### Reproduction-first audit

I am closing this PR unmerged. I did not first publish a complete real CLI/server transcript with exit codes and source/destination checks. The regression did not run at the exact PR head.

PR #1056, opened 2026-09-07
fix(ui): wait for document handoff before block scroll

### LINKED ISSUE #953: Cold navigation to a #blockRef deep link does not scroll to the block until reload
## What happens

Navigating to a `#blockRef` deep link as a **cold navigation** — typing or pasting `https://<site>/<doc>#<blockId>` into the address bar from another page — leaves the page at `scrollY 0`. The target block is never scrolled into view. Pressing F5 on the very same URL then lands it correctly.

Reproduced on web desktop and web mobile against a local stack with a 12-section test document:

- `localhost:3000/long-doc#5_yz1lxo` typed from another page → `scrollY 0`, target ~7500px below the viewport.
- Same URL reloaded → target lands 16px below the document top bar, as intended.
- `performance.navigation.type` was `navigate` in the failing case, the target element was present in the DOM, and waiting longer did not help.
- A hash change to a *different* block on a page that is already loaded works correctly (both typed and programmatic).

## Why it matters

This is exactly the "someone pastes a shared block link" path: a link to a specific block or a quoted comment target opens the document at the top with no indication of where the referenced content is.

## Where to look

`useBlockScroll` (`frontend/packages/ui/src/use-block-scroll.ts`) resolves the target and scrolls it. The failing case looks like the scroll losing a race with first render / hydration on a cold navigation — the effect likely runs before the block exists or before layout has settled, and unlike the reload path there is no native fragment scroll to fall back on (the browser only performs it for same-document fragment navigations, not for a client-rendered first paint).

## Not this

Not caused by the document top bar work in #950: the same behavior reproduces on commits before it, and the pinned-bar offset itself is correct (the reload path lands at the intended 16px). Filed separately at @horacioh's request.

ion-lion:
## Summary
- wait through the existing two-frame SSR-to-editor handoff before resolving a route block fragment
- keep the existing missing-target retry behavior after the live editor layout is ready
- add a deterministic regression covering the premature window-scroll fallback

Fixes #953.

## Validation
- `vitest --run src/__tests__/use-block-scroll.test.tsx` (3 passed)
- regression test fails against `origin/main` by observing the first-frame window scroll
- Prettier check passed for both changed files
- `git diff --check` passed
- `@shm/ui` typecheck was attempted twice but exceeded this sandbox's memory limit (Node OOM at 256 MB; process killed at larger heap limits); CI remains authoritative for typecheck

### DIFF
--- a/frontend/packages/ui/src/__tests__/use-block-scroll.test.tsx
+++ b/frontend/packages/ui/src/__tests__/use-block-scroll.test.tsx
@@ -1,7 +1,7 @@
 // @vitest-environment jsdom
 import {act} from 'react-dom/test-utils'
 import {createRoot, type Root} from 'react-dom/client'
-import {afterEach, beforeEach, describe, expect, it} from 'vitest'
+import {afterEach, beforeEach, describe, expect, it, vi} from 'vitest'
 import {useBlockScroll} from '../use-block-scroll'
 ;(globalThis as typeof globalThis & {IS_REACT_ACT_ENVIRONMENT?: boolean}).IS_REACT_ACT_ENVIRONMENT = true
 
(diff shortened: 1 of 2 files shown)


ion-lion:
### Reproduction-first audit

I am closing this PR unmerged. I did not first capture the cold block-link scrolling failure in a real browser. The JSDOM test manufactures the suspected timing state and is mechanism evidence, not reproduction.

PR #1059, opened 2026-09-07
fix(search): preserve matching document identity

### LINKED ISSUE #948: SearchEntities returns the wrong document id for text hits (always the space's newest document)
`SearchEntities` returns the correct title for a text hit but the wrong `id`: every hit within a space points at the **last document published in that space** rather than the document the text actually appears in.

Reproduced on a local daemon while runtime-testing the new Explore advanced search (horacioh/seed#20), across three freshly seeded spaces.

## Reproduction

Seed a space with several documents, then query the search API directly (no UI involved):

```
Search?query=oncall
→ entities[0].title = "Oncall Rotation Spec"        # correct
  entities[0].id    = hm://<uid>/notes/engelbart-history   # wrong: this is the space's most recently published doc
```

The word `oncall` appears only in `/specs/oncall-rotation`. The title proves the right document was matched; the `id` points elsewhere. This holds for document-level hits, not just block-level ones, and it happens for every space tested — the returned `id` is always the space's newest document.

## Why it matters

Anything that follows a search hit's `id` goes to the wrong resource:

- Clicking a search result navigates to an unrelated document (this is the same root cause as the long-standing "block result opens Document Not Found" behaviour in the existing header search).
- Anything that *joins* on the search result id breaks. Explore's advanced search intersects `SearchEntities` text hits with `QueryDocuments` attribute matches to answer queries like `Engelbart status="In Progress"`. With the wrong id, the text hit is attributed to a document that does not satisfy the attribute condition, so the intersection is empty and the query returns nothing even though a document plainly satisfies both halves. On a node exhibiting this, mixing words with attribute filters is unusable.

## Notes

- Not caused by horacioh/seed#20 — that branch only made it visible, since it is the first consumer to join search results against document identity. It reproduces through the raw API.
- Independently on the same node, full-text indexing was unreliable: `widget`, `augmentation`, `offsite` and `Zorblattix` returned nothing through the API even after `ForceReindex`, while other terms in the same documents matched. Possibly related, possibly a separate indexing gap — flagging it here rather than filing blind.

ion-lion:
## Summary
- constrain SearchEntities document-path resolution to generations that contain the matching FTS change
- retain current moved/republished-path selection among valid generations
- cover distinct title and body hits across multiple documents in one space

## Root cause
`current_document_resources` matched every live resource sharing a genesis, then selected the newest one. A genesis can back multiple document paths, so unrelated hits inherited the newest document ID. The generation already stores its reachable change set; intersecting it with the FTS row blob preserves the hit's document identity.

## Validation
- `gofmt` on both touched Go files
- `git diff --check`
- focused backend regression added; exact-head GitHub CI requested because the coordinator microVM's 128 MB `/tmp` cannot compile the backend test package without violating repository sandbox guidance

Closes #948

### DIFF
--- a/backend/api/entities/v1alpha/entities.go
+++ b/backend/api/entities/v1alpha/entities.go
@@ -371,6 +371,14 @@ effective_comment_resources AS (
   )
   WHERE rn = 1
 ),
+document_resource_candidates AS (
+  SELECT DISTINCT f.rowid, refs.resource
+  FROM fts_data f
+  CROSS JOIN structural_blobs refs INDEXED BY structural_blobs_by_genesis_blob
+  WHERE f.type IN ('title', 'document')
+  AND refs.genesis_blob = COALESCE(f.genesis_blob, f.blob_id)
+  AND refs.type = 'Ref'
+),
 current_document_resources AS (
   SELECT rowid, resource, metadata, heads, is_deleted
   FROM (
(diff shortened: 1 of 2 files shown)


ion-lion:
Final review found an additional generation edge despite the initial green CI: `resources.genesis_blob` can be updated by a delayed older-generation Ref, so using that mutable summary to discover candidates could hide the current generation.

Head `626f4121d` now derives candidate paths from immutable Ref history by genesis, retains the generation bitmap membership check, and adds `LateOldGenerationRef` coverage for tombstone → republish → delayed old Ref delivery. The PR remains draft while exact-head backend CI runs.

ion-lion:
### Reproduction-first audit

I am closing this PR unmerged. The upstream report contained an API mismatch, but I did not independently reproduce it against a named clean Seed baseline before editing, and no real-daemon exact-head transcript verifies the fix.

PR #1061, opened 2026-09-07
fix: discard empty failed discovery scopes

### LINKED ISSUE #957: Bug: arbitrary discovery paths persist empty rbsr_scope rows
## Problem

A request for a syntactically valid but nonexistent resource can leave a persistent `rbsr_scope` row. For example:

`hm://z6Mkvz9TgGtv9zsGsdrksfNk1ajbFancgHREJEz3Y2HsAVdk/wp-content/plugins/hellopress/wp_filemanager.php`

This appears capable of growing the table with one row per unique path, creating a storage and database-writer DoS risk.

## Reproduction

1. Run a daemon with the maintained RBSR index enabled.
2. Call the public web `DiscoveryStatus` endpoint with a valid account ID and a unique nonexistent path, or call `Entities.DiscoverEntity` directly with the equivalent `hm://` ID.
3. Wait for the discovery task to run and fail.
4. Query `rbsr_scope`:

```sql
SELECT iri, kind, materialized, last_access
FROM rbsr_scope
WHERE iri LIKE '%wp-content%';
```

The path is accepted and an exact or recursive scope row is materialized even when no matching resource exists.

## Findings

- `frontend/packages/shared/src/api-discovery-status.ts:29-39` calls `DiscoverEntity` for the status poll.
- `backend/api/entities/v1alpha/entities.go:163-203` validates only the IRI shape and version, then calls `TouchHotTask`; it does not check that the resource exists.
- `backend/hmnet/syncing/rbsr_index.go:171-215` treats a missing index key as cold and calls `resolveScope` followed by `materializeScope`.
- `backend/hmnet/syncing/rbsr_index.go:258-292` inserts the `(iri, kind)` row with `INSERT OR IGNORE`.
- `backend/storage/schema.sql:436-450` only enforces uniqueness for the same path and kind.
- `backend/hmnet/syncing/shadow_verify.go:49-55` has scope eviction disabled with `shadowVerifyEvictAfter = 0`, so idle rows are not removed.

The current gateway SSR path deliberately avoids starting discovery for a not-found HTML request. The browser shim then polls `DiscoveryStatus`, so a JS-capable crawler or direct API caller can still trigger this behavior.

## Expected behavior

Discovery of a resource that does not exist should not create a durable maintained scope, or failed/nonexistent discovery scopes should be bounded and cleaned up. Please also consider request rate limiting or another admission control for this path.

## Validation

`go test ./backend/api/entities/v1alpha ./backend/hmnet/syncing` passes.

ion-lion:
## Summary

- remove a maintained RBSR scope after a discovery wave returns no resource and the scope still contains no blobs
- keep populated scopes intact and make cleanup safe after caller cancellation with a bounded detached context
- avoid acquiring the SQLite writer on the successful/populated path by probing emptiness on a read connection first

Closes #957.

## Why

`DiscoverEntity` accepts arbitrary valid paths. The client-side maintained-index load materializes each path as an `rbsr_scope` row even when it resolves to an empty blob set. With eviction disabled, direct callers can grow the table without bound.

Cleanup runs only after the wave has finished and only when its result version is empty. The conditional delete rechecks that no `rbsr_item` exists in the writer transaction, so a scope populated concurrently is retained. Future attempts for a genuinely absent resource can recreate the cheap scope transiently.

## Validation

- `gofmt` clean
- `git diff --check` clean
- focused tests added for deleting an empty scope and retaining a populated scope
- local `go test ./backend/hmnet/syncing -run 'TestDeleteEmptyScope|TestRbsrIndex' -count=1` could not complete because this executor exposes a fixed 128 MB `/tmp`; compilation exhausted it. GitHub `test-go` and `lint-go` are authoritative.

### DIFF
--- a/backend/hmnet/syncing/discovery.go
+++ b/backend/hmnet/syncing/discovery.go
@@ -115,6 +115,23 @@ func (s *Service) DiscoverObjectWithProgress(ctx context.Context, entityID blob.
 		return "", fmt.Errorf("remote content discovery is disabled")
 	}
 
+	indexKey := DiscoveryKey{
+		IRI:       entityID,
+		Recursive: recursive,
+		DepthOne:  depthOne,
+		BlobTypes: BlobTypesString(blobTypes),
+	}
+	defer func() {
+		if resultVersion != "" {
+			return
+		}
+		cleanupCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 5*time.Second)
+		defer cancel()
+		if err := deleteEmptyScope(cleanupCtx, s.db, indexKey); err != nil {
+			s.log.Debug("RBSREmptyScopeCleanupFailed", zap.String("iri", string(entityID)), zap.Error(err))
+		}
+	}()
+
 	discoverStart := time.Now()
 	// This is the single entry point for every sync session — scheduler tasks,
 	// gRPC DiscoverEntity, and direct subscription calls all funnel here — so
--- a/backend/hmnet/syncing/rbsr_index.go
+++ b/backend/hmnet/syncing/rbsr_index.go
@@ -317,6 +317,47 @@ const qMaterializeReplace = `
 	JOIN blobs b INDEXED BY blobs_metadata ON b.id = rb.id
 	WHERE b.size >= 0;`
 
+var qEmptyScopeExists = dqb.Str(`
+	SELECT 1 FROM rbsr_scope s
+	WHERE s.iri = :iri
+	AND s.kind = :kind
+	AND NOT EXISTS (SELECT 1 FROM rbsr_item WHERE scope = s.id);`)
+
+var qDeleteEmptyScope = dqb.Str(`
+	DELETE FROM rbsr_scope
+	WHERE iri = :iri
+	AND kind = :kind
+	AND NOT EXISTS (SELECT 1 FROM rbsr_item WHERE scope = rbsr_scope.id);`)
+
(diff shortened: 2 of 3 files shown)


ion-lion:
Independent exact-head review found a lifecycle correctness risk before this is marked ready:

- `DiscoverObjectWithProgress` deletes on every `resultVersion == ""`, which includes cancellation, deadline, preemption, and transient failures that the scheduler intentionally retries. The scope may still be resumable/shared.
- recursive discovery can also materialize the derived `scopeDirStructure` (`scopeDirStructure` path around discovery.go:647), but cleanup only considers `indexKey`, so the reported unbounded-row case is not fully bounded.
- `NOT EXISTS rbsr_item` is not authoritative immediately after downloads because `MaintainRBSRIndex` is asynchronous (`backend/blob/index.go`); a useful scope can be deleted in the indexing window. Scope identity also ignores Version, so concurrent version tasks can share it.

Recommendation: move cleanup to scheduler task retirement (after retry/hot TTL ownership ends), cover all canonical scopes the task created, and retain scopes with another active task/subscription. Add a lifecycle regression that proves retriable/cancelled and overlapping owners retain the scope, then retirement of the last failed owner removes it. The helper tests currently validate SQL only, not safe invocation timing.

ion-lion:
### Reproduction-first audit

I am closing this PR unmerged. I did not first publish the real discovery failure and persisted database row on a clean baseline. Later review also found lifecycle defects in this patch; helper tests are not reproduction.

Issue #947, opened 2026-08-11
Daemon-maintained interaction summary: stop counting citations by enumerating them


ericvicenti:
## Problem

`InteractionSummary` derives `{citations, comments, changes, children, authorUids, blocks}` — a handful of small integers — by fetching the target's **entire citation list** through `ListCitations`. Each `ListCitations` call costs the daemon 0.3–2.5s regardless of page size, because `qListCitationsTpl` materializes the target's whole citation fan-out (the `changes` → `citing_blobs` UNION expanding each change by every Ref in its genesis chain) and sorts it in temp B-trees **before** applying `LIMIT`.

This was a primary cause of the 2026-08-11 production saturation (see `docs/daemon-saturation-incident.md`): unbounded chains of these calls from `/api/InteractionSummary` held the 12-connection SQLite read pool and convoyed every other read. #946 capped every client-side enumeration to one page as a stopgap, at the cost of under-reporting counts for documents with >500 citations (two exist in production today, ~6.4k citations each).

## Proposed fix

Serve the summary from the daemon with one bounded aggregation, the way `children_count` already works (`qDocumentsOuterColumns` in `backend/api/documents/v3alpha/documents.go` does the cheap indexed thing for directories).

Two possible shapes:

1. **Extend `ActivitySummary`** with `citation_count` (and optionally `latest_citation_time`), computed by an indexed `COUNT` over `resource_links` by target — enough for header/card counts, and lets counts return to SSR if desired.
2. **New `GetInteractionSummary` RPC** that also returns the per-block `{citations, comments}` breakdown (`GROUP BY` on `resource_links.extra_attrs->>'f'`) and distinct author list, replacing the client-side `calculateInteractionSummary` entirely, including dedup semantics (`deduplicateCitations`).

Option 2 removes the enumeration from every consumer (web margin markers, embed cards, desktop) and makes the result cacheable server-side. Option 1 is a smaller first step.

Either way the aggregation must be **bounded**: no per-row genesis-chain expansion, index seeks only.

## Also worth doing while in there

- `qListCitationsTpl` pays the full fan-out cost even for one page; if the citations panel is to show >500 citations again, pagination needs to become genuinely incremental.
- Per-RPC deadlines / in-flight caps and Remix client-disconnect propagation are the systemic guards (incident follow-ups 4–5).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1063, opened 2026-09-07
perf: expose indexed citation counts in document info

ion-lion:
## Summary
- add `ActivitySummary.citation_count` to the daemon API and generated Go/TypeScript bindings
- compute it as distinct citing document roots (`coalesce(genesis_blob, id)`) from `resource_links_by_target`
- cover multiple links and multiple changes from one source without inflating the count

This is the smallest backend slice of #947. It makes an index-bounded document citation count available without `ListCitations`; this draft will stay open while redirect, deletion, and public-visibility semantics are covered before consumer migration.

## Query evidence
Synthetic production-sized fan-out: 6,400 inbound links within 106,400 total links, 3,200 distinct source documents.

- `SEARCH rl USING COVERING INDEX resource_links_by_target (target=?)`
- `SEARCH sb USING INTEGER PRIMARY KEY (rowid=?)`
- median 2.480 ms, p99 5.265 ms over 250 warm runs

The only temporary structure is the bounded `count(DISTINCT)` set for the target's inbound links; the query does not scan unrelated links.

## Validation
- generated Go and TypeScript checksums verified
- `git diff --check`
- focused Go test added (`TestDocumentInfoCitationCount`)
- CI informed two identity fixes: Change rows do not carry source `resource`, and initial Changes have no `genesis_blob`; head `3966744af` uses `coalesce(genesis_blob, id)`
- exact-head `test-go`, `lint-go`, frontend unit/lint/typecheck, and editor E2E checks passed
- local execution could not finish within the sandbox's 300-second cold Go compile limit; exact-head CI is authoritative

### Source lifecycle semantics

Head `c73f00143` resolves each citing Change through its current Ref-backed resource identity, collapses moved source paths, and excludes tombstoned source documents. `TestDocumentInfoCitationCount` now covers both transitions. All exact-head frontend, Go lint, and Test Go checks passed. Local cold-build resource limits prevented a focused result, so CI is the authoritative test evidence.

### Public-only visibility semantics

Head `fda01a181` preserves `ListCitations` visibility rules in the aggregate: anonymous callers on public-only nodes count only public source Changes, while authenticated target-account root writers may also count private citations. The SQL embeds only the already-verified caller principal as a hex blob literal, avoiding changes to positional query bindings. `TestDocumentInfoCitationCount` covers both anonymous and owner views. Local focused execution again exceeded the sandbox's 300-second cold compile window; exact-head CI is running.

### Consumer migration

Head `8129b70f5` uses `ActivitySummary.citation_count` for the top-level InteractionSummary citation total instead of deriving that total from the bounded citation page. The citation page remains temporarily for per-block markers, comment counts, and author IDs; removing that RPC requires daemon-side aggregates for those payload fields. Focused shared tests passed (12/12) and formatting passed.

### Exact-head CI note

At `8129b70f5`, shared/web/desktop unit tests, lint, typecheck, editor E2E, Go lint, and the aggregate check passed. The documents v3alpha package (including the new visibility regression) passed inside Test Go. That job is red only because six daemon tests fail during unrelated private-document setup with `private document creation is disabled`, reproduced on two consecutive heads; Ion lacks upstream Actions admin permission to rerun it.

## Latest aggregate-RPC slice

Head `6349e7a20` adds bounded `GetInteractionSummary`, grouped from `resource_links_by_target`, and removes InteractionSummary's remaining `ListCitations` call. It returns complete document/comment totals, per-fragment counts, and distinct author IDs while preserving public-only, move, tombstone, and comment-deletion semantics without genesis-chain expansion.

Validation at this head:
- focused shared tests: 12/12 passed
- generated Go bindings compile
- full frontend typecheck and pre-push checks passed on the preceding aggregate head; exact-head CI is running
- local focused backend compilation exceeded the constrained executor's 300-second budget; exact package CI is authoritative


ion-lion:
Independent exact-head CI review for `d05deb236357d2b6cf659286dd31df5e4a8dc8d7`:

- `seed/backend/api/documents/v3alpha` passed in Test Go (`2.013s`), including the corrected interaction-summary/comment regression.
- The job-level failure is isolated to six pre-existing `seed/backend/daemon` tests; every failure occurs during setup with `private document creation is disabled`.
- All other reported exact-head checks, including Go lint, frontend unit/typecheck/lint, and the aggregate check, are green.

This removes the red Test Go badge as a blocker for reviewing the touched package; the daemon fixture/configuration failure remains unrelated to this PR.

ion-lion:
Exact-head follow-up at `5908773ee23f9f82f432878701bfcc26e8e37a4d` closes four semantic gaps found during adversarial review:

- follows bounded redirect ancestors so interactions survive target moves;
- scopes comment revisions/deduplication by `(author, TSID)`;
- restricts current Ref author selection to visible Refs for public-only callers;
- returns an empty aggregate for an existing uncited comment instead of `NotFound`.

Added regressions for moved targets and uncited comments. Exact-head `seed/backend/api/documents/v3alpha` passed (2.524s), as did Go lint and all frontend checks. The job-level Go failure remains isolated to the same six unrelated daemon fixture tests failing private-document setup.

ion-lion:
### Reproduction-first audit

I am closing this PR unmerged. The incident record motivated this work, but I did not first run and publish the exact baseline acceptance/benchmark myself. The current head also lacks raw SHA-bound post-fix benchmark and concurrent-RPC evidence.

Issue #779, opened 2026-06-15
Bug: rework comment counts on app


horacioh:
sometimes we are not counting the full thread comment count on summaries. we need to rework this so we get the same results everywhere.

PR #1066, opened 2026-09-07
perf(backend): maintain comment activity by genesis, index feed ordering

juligasa:
## Why

A 100s CPU profile of the production gateway shows `sqlite3_step` using **148 of 193 CPU-seconds — 77% of everything the daemon does**. Three read endpoints account for most of it, and all three recompute, on every request, data that doesn't change between requests:

| Method | CPU | Share |
|---|---|---|
| `Documents.ListDirectory` | 67.4s | 35% |
| `ActivityFeed.ListEvents` | 19.3s | 10% |
| `Documents.GetAccount` | 18.5s | 9.6% |

## What this does

**1. Comment activity is maintained, not derived** (`backend/blob/comment_stats.go`). Two tables settle at index time what the listings used to recompute per request: `comment_live` (which blob is the live version of each comment TSID) and `document_comment_stats`. The listings read them with one indexed point lookup.

**2. One index for the feed**, `structural_blobs_by_ts (ts, id)`. Neither feed query could page through an index, so SQLite materialised ~40k rows — each carrying the whole `extra_attrs` JSONB — into temp b-trees to return 30. The main query's `SELECT DISTINCT` is also dropped: every join is many-to-one on a primary key, so it removed nothing (37912 rows either way).

**3. Space comment totals are recomputed, not accumulated.** `spaces.comment_count` had drifted to **5062 against 11964 real** — delta arithmetic plus an early return that silently skipped updates. Now exact.

**4. Comment activity is keyed by genesis, not by redirects** — from Alex's review.

## Results

Measured on a 6.2 GB copy of production, with the daemon's own vendored SQLite:

| Query | Before | After |
|---|---|---|
| `ListDirectory`, busiest space | 52ms | **2.3ms** |
| `ListEvents` main query | 175ms | **~0.05ms** |
| `ListEvents` mentions query | 250ms | **~0.06ms** |
| `getDocumentInfo` | 3.0ms | **0.11ms** |

Reindex on that database also got *faster*, 1m50s → 1m39s, because the redirect walks are gone.

## The genesis change, and why it matters

A comment belongs to a document, and a document's identity is its genesis — not the path it currently sits at. That makes both of the things reviewers wanted true at once, where the old redirect-graph rule could only manage one:

- a document that **moves** keeps its comments at every path in the chain, because moving republishes the same genesis;
- a redirect between two **unrelated** documents transfers nothing, because their genesises differ.

The second was already wrong in production. Nothing in any of the four redirect walks checked document identity — `RedirectTarget` is `{Space, Path, Republish}`, a location. On the production copy:

| | |
|---|---|
| Redirect edges | 1766 |
| Same genesis (real moves) | 1759 |
| **Different genesis** | **6** |
| …of those, carrying comments | 2 |

`/tech-talks` redirects at `/tech` with a different genesis, and was pushing **293 comments** onto a document they were never written against. After this change `/tech` reports 1 and `/tech-talks` keeps its 293.

`ListComments` changes in lockstep — otherwise the count and the list disagree about what a document's comments are, which is the complaint in #779. Querying by a path the document moved away from still works, because that path's own latest generation carries the same genesis.

## Worth a reviewer's attention

**One behaviour change.** A comment whose target changes we *hold but haven't indexed yet* is now stashed and retried, instead of being indexed with no attribution. Without it, reindexing attributed only **6602 of 12659** comments — a comment replayed before its target was skipped and nothing revisited it. With it, all **11964** are attributed and space totals agree exactly.

A comment whose target changes we **don't hold at all** (`BlobsSize < 0`) is still indexed and left out of the counts, so nothing gets hidden waiting for a blob that may never arrive.

**It deletes more than it adds.** Both recursive redirect walks are gone, along with the end-of-reindex rebuild they required and the recursive CTE in `ListComments`.

## Tests

- `TestCommentActivityFollowsTheDocument` — pins both halves: a twice-moved document keeps all four of its comments, and a redirect to an unrelated document transfers none. Also checks `ListComments` agrees with the counts, that old paths still resolve, and that a reindex reproduces the same numbers.
- `TestCommentStatsQueriesSeek` / `TestFeedQueriesUseIndexedOrdering` — plan guards. These failure modes are silent: the wrong plan still returns the right answer, just slowly. One of these already caught a full-table scan I'd written by accident.
- Pre-existing `TestCommentCount_*`, `TestListComments_AfterDocumentMove` and `TestRedirect` pass unchanged.

## Verify

```
go test ./backend/...
```

For the production numbers: point a daemon at a copy of a real database, let it migrate and reindex, then compare `document_comment_stats` against `comment_live` grouped by genesis.

### DIFF
--- a/backend/api/activity/v1alpha/activity.go
+++ b/backend/api/activity/v1alpha/activity.go
@@ -105,6 +105,52 @@ func (srv *Server) RegisterServer(rpc grpc.ServiceRegistrar) {
 	activity.RegisterSubscriptionsServer(rpc, srv)
 }
 
+// buildMainEventsQuery assembles the feed's main blob-events query. filtersStr
+// is the caller's already-rendered filter prefix (each term ending in " AND ",
+// possibly empty); orderByObserved selects the observed-time cursor column
+// instead of the claimed-time one.
+//
+// No DISTINCT: every join here is many-to-one on a primary key (blobs.id is
+// structural_blobs' own primary key, and public_keys.id and resources.id are
+// the other two), so one structural_blobs row can only ever produce one output
+// row. The DISTINCT that used to be here removed nothing — verified on a
+// production database, 37912 rows either way — while costing a temp b-tree
+// over every row that survived the WHERE: ~100ms of the query's 175ms.
+//
+// The remaining cost was the ORDER BY. Neither cursor column had an index that
+// could serve the ordering, so SQLite materialized all ~40k surviving rows,
+// each carrying the whole extra_attrs JSONB, and then discarded all but the
+// page. structural_blobs_by_ts (schema.sql) fixes that for the claimed-time
+// order; the observed-time order sorts by structural_blobs.id, which is the
+// table's own primary key. Both now terminate at the LIMIT.
+//
+// It's a function rather than inline SQL so TestFeedQueriesUseIndexedOrdering
+// can assert the plan of the exact statement this runs.
+func buildMainEventsQuery(filtersStr string, orderByObserved bool) string {
+	var (
+		selectStr            = "SELECT " + storage.BlobsID + ", " + storage.StructuralBlobsType + ", " + storage.PublicKeysPrincipal + ", " + storage.ResourcesIRI + ", " + storage.StructuralBlobsTs + ", " + storage.BlobsInsertTime + ", " + storage.BlobsMultihash + ", " + storage.BlobsCodec + ", " + "structural_blobs.extra_attrs->>'tsid' AS tsid" + ", " + "structural_blobs.extra_attrs" + ", " + storage.StructuralBlobsGenesisBlob
+		tableStr             = "FROM " + storage.T_StructuralBlobs
+		joinIDStr            = "JOIN " + storage.Blobs.String() + " ON " + storage.BlobsID.String() + "=" + storage.StructuralBlobsID.String()
+		joinpkStr            = "JOIN " + storage.PublicKeys.String() + " ON " + storage.StructuralBlobsAuthor.String() + "=" + storage.PublicKeysID.String()
+		leftjoinResourcesStr = "LEFT JOIN " + storage.Resources.String() + " ON " + storage.StructuralBlobsResource.String() + "=" + storage.ResourcesID.String()
+
+		mainCursorColumn = storage.StructuralBlobsTs.String()
+	)
+	if orderByObserved {
+		mainCursorColumn = storage.StructuralBlobsID.String()
+	}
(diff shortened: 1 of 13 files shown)


burdiyan:
Shall we not fix `spaces.comment_count` too here then? @juligasa

burdiyan:
Wait, why comment counts are affected by redirects at all? I think we should count comments based on the genesis CID of the document that's being targeted by the comment. Then regardless of where the document is placed we can count and list comments properly.

But if `/a` redirects to `/b` and **they may be totally different documents**, then why would we aggregate those comment counts?

ericvicenti:
Bob comments on doc A, doc A gets moved to location B. Now, when I see location B show up in the list of documents, it needs to show Bob's comment in the count. And of course that comment must continue appearing at location B.

burdiyan:
Redirect doesn't always mean moved. Redirect can point anywhere. So only when the genesis is the same we can infer that they are the same documents. Which is what I'm talking about.

juligasa:
> Shall we not fix `spaces.comment_count` too here then? @juligasa

Good point, I added it. It does not use redirects at all, so it is safe in this PR. It was counting 5062 comments, but the real number is 11964. Same delta problem, plus an early return that sometimes skipped the update.

juligasa:
> Wait, why comment counts are affected by redirects at all? I think we should count comments based on the genesis CID of the document that's being targeted by the comment. Then regardless of where the document is placed we can count and list comments properly.
> 
> But if `/a` redirects to `/b` and **they may be totally different documents**, then why would we aggregate those comment counts?

You are right. And we already have this problem in main.
/tech-talks redirects to /tech, but they have a different genesis. So /tech shows 293 comments that belong to another document.
Nothing checks the document identity. RedirectTarget is only {Space, Path, Republish}, and all four walks compare only redirect_iri = iri.
Two problems if we fix it: comments do not store the genesis of the target, and the simple check src.genesis == tgt.genesis does not work, because indexing a redirect Ref writes the genesis of the target into the source.

burdiyan:
@juligasa 

> comments do not store the genesis of the target

Comments always carry CIDs of the version of the target. So we can look up their genesis, because changes do carry the genesis. And if those changes not available, then we should probably stash the comment altogether because there's nowhere to show it.

And from now we don't care about redirects at all. Whatever the visible document shows as genesis, we can aggregate comment counts per genesis.

juligasa:
> @juligasa
> 
> > comments do not store the genesis of the target
> 
> Comments always carry CIDs of the version of the target. So we can look up their genesis, because changes do carry the genesis. And if those changes not available, then we should probably stash the comment altogether because there's nowhere to show it.
> 
> And from now we don't care about redirects at all. Whatever the visible document shows as genesis, we can aggregate comment counts per genesis.

Done, it is keyed by genesis now and all the redirect walks are gone.
/tech went from 294 comments to 1, and /tech-talks keeps its 293. ListComments changed too, so the count and the list always agree.

burdiyan:
LGTM

burdiyan (approved):

PR #1068, opened 2026-09-07
fix(editor): preserve rich paste with remote images

### LINKED ISSUE #735: Pasting rich text from some webpages fails when copied HTML includes cross-origin images
## Summary

Pasting selected content from some webpages into the desktop app fails when the copied HTML includes remote image URLs that are fetched from the renderer/browser context and blocked by CORS.

Example page:  
https://xanadu.com/zigzag/

When copying content from that page and pasting into the app, the paste flow appears to try to fetch an image referenced in the copied HTML, hits a CORS error, and aborts the paste.

## Steps to reproduce

1. Open https://xanadu.com/zigzag/
2. Select visible page content
3. Copy
4. Paste into the desktop app

## Actual behavior

Paste fails with errors like:

- `Access to fetch at 'https://xanadu.com/zigzag/zzlogo.jpg' from origin 'http://localhost:17654' has been blocked by CORS policy`
- `Error processing pasted HTML image: TypeError: Failed to fetch (xanadu.com)`

## Expected behavior

The app should support pasting rich HTML that includes remote images.

Ideally:

1. detect remote image references in pasted HTML
2. fetch/import those images via a native/main-process/backend network path that is not blocked by browser CORS
3. store/embed them locally
4. preserve the images in the pasted result

If image import still fails for any reason, the paste should degrade gracefully and still paste the text/HTML content rather than failing entirely.

## Why this should be fixed in-app

This does not seem like a source-site problem. The webpage is copyable, and the image URL is publicly accessible. The failure happens because the app is trying to fetch the image from the renderer/browser environment, where normal web CORS restrictions apply.

Example failing asset:

- `https://xanadu.com/zigzag/zzlogo.jpg`

The asset loads normally, but does not send permissive `Access-Control-Allow-Origin` headers, so renderer-side fetch fails.

## Suggested fix

Instead of relying on browser `fetch` during HTML paste for remote media:

- move remote image fetching to a native/main-process/backend layer
- import the image bytes there
- rewrite pasted HTML/content to point at the imported local asset
- treat image import failures as non-fatal

## Why this matters

Users expect copy/paste from ordinary webpages to work, including embedded images when possible. Sites that do not opt into cross-origin image reads are common, so renderer-side fetching will keep breaking valid paste flows.

## Notes

This issue is not just about avoiding images or stripping them out. The better fix is to make remote image import work in the app architecture, with graceful fallback only when import is genuinely impossible.

ion-lion:
## Summary
- stop the local-media plugin from consuming rich HTML just because it contains an image
- let the editor schema paste the complete HTML, preserving adjacent text and every image
- keep direct clipboard image/file handling unchanged
- avoid renderer-side remote image fetches; desktop image blocks use the existing main-process `importWebFile` path and failed imports remain non-fatal

Fixes #735.

## Root cause
The plugin extracted the first `img[src]`, returned `true` immediately, and fetched that URL in the renderer. A CORS failure then only logged an error, after the whole paste had already been consumed.

## Validation
- rich HTML with a remote image is not claimed and renderer `fetch` is never called
- direct clipboard image files still use the attachment handler and insert an image node
- external image attributes remain accepted by the image schema
- native import success rewrites the image to `ipfs://`; rejection leaves the original remote image intact
- exact-head frontend CI is the test authority for the current head

## Security boundary
This adds no IPC endpoint or renderer capability: desktop document editors already receive the callable `importWebFile` procedure, and image blocks already invoke it for external URLs. Rich-HTML paste now reaches that existing path after an explicit user paste instead of auto-fetching the same URL in the renderer. The generic importer's pre-existing URL/redirect/size policy remains a separate hardening concern and is called out rather than obscured here.

## Scope
Existing direct clipboard-file paste and web/comment attachment handling remain in place.

### DIFF
--- a/frontend/apps/desktop/src/app-web-importing.ts
+++ b/frontend/apps/desktop/src/app-web-importing.ts
@@ -4,15 +4,19 @@ import {DAEMON_FILE_UPLOAD_URL} from '@shm/shared/constants'
 import {htmlToBlocks} from '@shm/shared/html-to-blocks'
 import * as cheerio from 'cheerio'
 import {app, dialog} from 'electron'
+import {lookup} from 'dns/promises'
 import {readFile, writeFile} from 'fs/promises'
 import http from 'http'
 import https from 'https'
 import {nanoid} from 'nanoid'
+import {isIP} from 'net'
+import type {LookupFunction} from 'net'
 import {join} from 'path'
 import z from 'zod'
 import {getSigner, seedClient} from './app-client'
 import {userDataPath} from './app-paths'
 import {t} from './app-trpc'
+import {isPublicIPAddress, normalizeIPHostname} from './remote-file-security'
 import {PostsFile, ScrapeStatus, scrapeUrl} from './web-scraper'
 import {serializeImportFile} from './wxr-crypto'
 import {cancelWXRImport, getImportStatus, hasActiveImport, resumeWXRImport, startWXRImport} from './wxr-import'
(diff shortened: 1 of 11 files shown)


ion-lion:
Blocking security/data-preservation review at `79de5bb35a0635dc9917d6e811a8175a7337507e`:

1. Rich clipboard payloads that expose both `text/html` and an image `File` are still consumed by the direct-file branch before normal HTML parsing. That drops adjacent text and additional images. Prefer non-empty rich HTML before file flavors and keep direct-file handling for standalone file pastes.
2. Letting arbitrary pasted `` values reach `ImageBlock` newly makes clipboard content invoke desktop `importWebFile` automatically. The main-process downloader currently permits arbitrary protocols/hosts via HTTP fallback, private/loopback/link-local destinations, unlimited redirects, no timeout, and unbounded buffering. This turns untrusted clipboard HTML into SSRF/resource-exhaustion input. Keep this PR in draft until the importer is constrained to HTTP(S), validates every DNS result and redirect against private networks, caps redirects/bytes/time, and rejects credentials.
3. Web document editors do not inject `importWebFile`, so this change preserves remote URLs rather than durable attachments; web comment import also needs verification that `mediaRef` survives.

I validated a bounded hardening approach locally (DNS-pinned public address per request, redirect revalidation, 5-hop/20 MB/15 s caps) but did not push after detecting another coordinator owns the mutable branch. Exact tests should cover mixed HTML/file flavors, private-address rejection (including redirects), size/time caps, desktop success without CORS, and non-fatal failure.

ion-lion:
Pushed hardening commit `eff985e613503c219467fb0df766d3544985f72a` after adversarial review. Remote media imports now pin validated DNS results, reject non-global/special IPv4 and IPv6 destinations on every redirect, enforce one absolute 15s deadline and 20 MB body cap, bound concurrent downloads, and preserve mixed HTML+video/file clipboard attachments. Local pure address-policy suite: 34/34 passed; exact-head frontend CI is the remaining authority for workspace typecheck and editor tests.

Issue #856, opened 2026-07-09
Bug: AGENTS: If I add a custom Model provider, I'm not able to create the agent because the model selector fails


horacioh:
## Problem

When adding a custom model provider, there's no way to create an agent because the model selector fails

<img  alt="Image" src="https://github.com/user-attachments/assets/e0887e96-ca17-4cf4-82f1-c842ae78718e" />

## Steps to Reproduce

1. open the agents panel
2. add a new provider
3. use Custom
4. create the provider
5. create a new agent
6. select the custom provider
7. see the model selector error

## Expected Behavior

either the model selector should be populated or I should be able to type the correct model

ericvicenti:
Thanks for filing! Agree that you should be able to type the model as a fallback.

What LLM server are you using? Ollama?

horacioh:
Flux.2 I'm building two agents:
- art director agent to create a prompt for an abstract image
- a image generator agent to create it

the second agent I can't create

ericvicenti:
Sorry if the UX is not clear about this: the "custom model provider" requires an OpenAI-compatible API, otherwise it doesn't know how to interact with the LLM server

I'm not sure, but it seems that Flux.2 has a totally different API. So I'm not sure that your workflow is possible with the current system.

PR #1071, opened 2026-09-07
Allow manual model IDs for Custom providers

ion-lion:
## Summary
- allow OpenAI-compatible Custom providers to select a typed, case-sensitive model ID when discovery fails, is empty, or has no exact match
- keep catalog-only providers restricted to discovered models
- retain custom model IDs through creation and session model switching while clearing stale values on provider changes

## Scope
This removes the UI catalog dead end for arbitrary model IDs. It does not add support for provider-specific image-generation APIs such as Flux.2.

## Tests
- focused provider-model selector Vitest: 5/5 passed on rebased head
- Prettier check passed for all changed files
- `git diff --check` passed

Rebased onto current `main` on 2026-09-09. The earlier `use-media` import fix commit was dropped during the rebase because the same change already landed on `main` via 4ad8d8d0f.

Addresses #856.

### DIFF
--- a/frontend/packages/ui/src/__tests__/provider-model-select.test.tsx
+++ b/frontend/packages/ui/src/__tests__/provider-model-select.test.tsx
@@ -1,0 +1,136 @@
+// @vitest-environment jsdom
+import React from 'react'
+import {createRoot, type Root} from 'react-dom/client'
+import {act} from 'react-dom/test-utils'
+import {afterEach, beforeEach, describe, expect, it, vi} from 'vitest'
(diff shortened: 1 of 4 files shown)


ion-lion:
Ready for review. Exact-head frontend CI at `6c8099d3` passed lint, desktop/web/shared unit tests, and editor E2E. Typecheck is red only on current-main test typing defects: the PR merge run reports `agent-session-rows.test.ts` and missing `serverUrl` props in `run-work.test.tsx`, while base `f5591a96` has the same failing Typecheck check ([base job](https://github.com/seed-hypermedia/seed/actions/runs/34162579625/job/101867243405)). The focused custom-provider selector regressions pass 5/5.

PR #1072, opened 2026-09-07
fix(frontend): repair agent test Typecheck failures

ion-lion:
## Summary
- preserve the discriminated-union narrowing in the agent session row timing regression by storing the final row before reading `message`
- provide the required agent server URL in all `RunWorkHierarchy` test fixtures

## Why
Current `main` fails the frontend Typecheck job on these two test-only typing defects. The base failure is visible in [this Typecheck job](https://github.com/seed-hypermedia/seed/actions/runs/34162579625/job/101867243405), and an exact merge run for #1071 isolated the same diagnostics.

## Validation
- `git diff --check`
- Prettier check on both touched files
- Full desktop Typecheck was attempted locally but the constrained executor was killed at its memory limit; exact-head frontend CI is the deciding validation.
- Focused Vitest under the executor's Bun fallback could not initialize the repository's Zod module (`z.object` undefined), so it does not represent a test assertion failure.

## Follow-up exact-head CI repair

The first exact-head run exposed two additional current-main `@shm/ui` diagnostics after the original test-only errors were cleared. Commit `7c8d7e6d4` removes the now-unused `profileId` local and switches the package-internal `useMedia` import to its resolvable relative path. `git diff --check` passes; exact-head CI is the decisive Typecheck evidence in this constrained executor.

### DIFF
--- a/frontend/apps/desktop/src/__tests__/agent-session-rows.test.ts
+++ b/frontend/apps/desktop/src/__tests__/agent-session-rows.test.ts
@@ -285,8 +285,8 @@ describe('buildAgentSessionChatRows step timing', () => {
     expect(starts).toEqual([1_700_000_000_030, 1_700_000_000_030])
     // Without the runs the wait still reads as deliberation, as it always did.
     const withoutRuns = buildAgentSessionChatRows(events, CONTEXT)
-    const part =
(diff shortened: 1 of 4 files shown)


ericvicenti:
Closing without merging: all four fixes here already landed on main via 4ad8d8d0f (ci: make main green) and 5bfe2b520 (test(desktop): bring agent run tests up to date). Main Typecheck is green on recent runs, and this branch is now conflicting.

PR #1073, opened 2026-09-07
fix(ui): retain document title after publish

### LINKED ISSUE #558: Document Title disappears.
Repro:
1. Open a document.
2. Change the document and publish it.
3. Open comments
4. Back to content
5. See the error, the title has disappeared.
It only happens sometime, except when I am demoing that happens all the time. 

Are you able to reproduce the bug? Come on people! I know you have experience it.. :)
@ericvicenti @horacioh @iskaktoltay

ion-lion:
## Summary
- make the editable header consume the document machine’s canonical effective metadata
- preserve the freshly published title/summary while the resource query refetches
- add focused regression coverage for published metadata plus a staged overlay

Fixes #558.

## Evidence
- current main clears `context.metadata` before storing the publish result in `context.document`, while the editable header only read the cleared overlay and potentially stale query data
- locked Prettier 3.1.1 check passes for all changed files
- `git diff --check` passes
- the constrained executor could not complete a targeted workspace dependency link; exact-head frontend CI is the decisive unit/typecheck evidence
## Exact-head CI

Run 34171842367 passed Lint, the new shared unit regression, all shared/desktop/web unit jobs, and editor E2E. Typecheck reached only the two pre-existing current-main diagnostics repaired by ready PR #1072; this branch introduced no additional diagnostic.

### DIFF
--- a/frontend/packages/shared/src/models/__tests__/document-machine.test.ts
+++ b/frontend/packages/shared/src/models/__tests__/document-machine.test.ts
@@ -5,6 +5,7 @@ import {
   deriveDocumentType,
   documentMachine,
   DocumentMachineInput,
+  getEffectiveDocumentMetadata,
   PushDocumentInput,
   retargetQueryBlockIncludesForPublish,
   WriteDraftOutput,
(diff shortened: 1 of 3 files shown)


ericvicenti:
oh man that issue is so old I'm not even sure its still a bug.

@horacioh maybe we should just close this and the issue? does it repro anymore?

horacioh:
this is fixed already

PR #1078, opened 2026-09-09
perf(agents): list sessions separately from GetAgent, paginated; cache negative per-session lookups

ericvicenti:
## Why

On agents-stable (agentic.seed.hyper.media) the Bun main thread sat at 100% CPU and session/run pages "never finished loading". Profiling (inspector `ScriptProfiler` + strace + tcpdump) showed:

- one client IP sending ~15 req/s, all `GetAgent` + `ListSessions`, driven by the UI refetching agent detail on every websocket session event;
- `GetAgent` returning **every** session of the agent (one agent has 509) with a trigger-context lookup and two continuation-link lookups per row — ~1,500 statements per call — all on the single JS thread;
- responses queued past Bun's 10s idle timeout, so Caddy logged 502s and the client retried into the same queue.

## What

**Server**
- `GetAgent` returns `sessionCount` (top-level) and no longer carries `sessions`. Sessions come from the existing paginated `ListSessions {agentId, includeChildren: false, cursor}` (already admits public-read agents for viewers).
- Trigger-context and continuation-link caches now store negative results (most sessions have neither). A new `#forgetSessionDerived` evicts a session's entries at every write that attaches a firing to a session and when a continuation edge is inserted, so a cached "nothing" never outlives the write.
- Protocol type + `docs/plans/agents.md` updated; tests assert through `ListSessions`.

**UI**
- New `useAgentSessions` (infinite query, 50/page, top-level only) in `@shm/ui/agents/models`.
- Agent detail page (desktop + web): sessions tab uses it with a **Load more** button; header count and move-agent dialog use `sessionCount`.
- Session page header: `sessionCount`.
- Assistant panel sidebar (`useSpaceAgents`): first page of sessions per space agent, fetched in parallel with the agent.
- Mobile `AgentScreen`: paginated list + Load more; pull-to-refresh covers both queries. Dev mock server updated.
- Optimistic cache patchers (rename, delete, websocket updates) handle the paged shape and the sidebar copy.

## Verification

| Check | Result |
|---|---|
| `agents` typecheck | clean |
| `agents` api-service + main tests | 142 pass |
| frontend typecheck (ui, desktop, web, cli, shared, editor, explore, notify, mobile) | clean |
| `@shm/ui` vitest, full package | 461 pass |
| mobile jest `models-import.test.ts` | fails on `main` too (transform SyntaxError), environmental |

## Compatibility note

Older desktop builds that read `sessions` from `GetAgentResponse` will get `undefined`. The detail tab is guarded; the old session-page header does `sessions.length` and would throw on that page until updated. If that matters for the rollout, the server can send `sessions: []` for one release.

## Not in this PR

- Prepared-statement caching for bun:sqlite (separate branch `perf/sqlite-stmt-cache`).
- Explicit `idleTimeout` on `Bun.serve`.
- The websocket handler still invalidates agent detail on session events; `GetAgent` is now two queries and its agent row carries the unread-activity columns (#1077), so that refetch looks intentional.

https://claude.ai/code/session_01UggBMiKG7c124E4nSgynEj

### DIFF
--- a/agents/docs/websocket-subscriptions.md
+++ b/agents/docs/websocket-subscriptions.md
@@ -59,7 +59,7 @@ type AgentWSEvent =
   | {
       _: 'change'
       key: `account/${string}`
-      value: {reason: string; agentId?: string; sessionId?: string; activity?: AgentActivity}
+      value: {reason: string; agentId?: string; sessionId?: string; activity?: AgentActivity; session?: SessionInfo}
     }
   | {_: 'change'; key: `runs/${string}`; value: RunInfo}
   | {_: 'append'; key: `runs/${string}`; runId: string; seq: number; entry: Record<string, unknown>; createdAt: number}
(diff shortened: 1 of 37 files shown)


ericvicenti:
Merged `origin/perf/sqlite-stmt-cache` (#1079) into this branch so both can be tested together. The merged suite caught a real bug in this PR: the continuation-links cache keyed on `${account}\0${session}` while the new eviction used a space, so a fresh continuation edge stayed hidden for the TTL. Fixed via a shared `sessionDerivedCacheKey`. Full agents suite: 485 pass.

PR #1084, opened 2026-09-09
perf(backend): index each block once per change; single-row fts move lookup

juligasa:
## Problem

A production CPU profile of the migration reindex (100 s window, 169.8 s CPU) spends **95 s of the reindex's 96.8 s in a single SELECT**: the raw-content lookup that the `OpMoveBlocks` branch of `indexChange` runs for every block entry of every move op. All fts inserts together take 0.7 s; everything else in indexing takes 1.75 s. The query plan is fine on both the system SQLite and the vendored 3.45.2; the cost is volume, and it is quadratic:

1. A change can list the same block in move ops many times. The block tree's `Commit` traversal yields a subtree once per historical position of its ancestors (no visited set in the BFS at `crdt_block_tree.go:370`, and old positions are never removed from the parent lists). Before 88d018f9f those repeats were batched into a single op, which is the shape in prod: one op moving the same block 1962 times. Frontend clients also build `MoveBlocks` ops directly without docmodel. The `Commit` duplicate yield will get its own PR; the indexer has to tolerate repeated ids regardless.
2. Each entry scanned every fts row the block already had for that document and then inserted one more duplicate (`INSERT OR REPLACE INTO fts` never replaces, and `fts_index` keys on a fresh rowid). The worst group in the April prod backup has 2044 rows for one 2-character block, 2029 of them from one change.
3. 98% of the lookups were wasted: across the backup, 53,538 of 62,608 lookups are for blocks the same change also replaces, and the `ReplaceBlock` op writes the final content itself. The move row was a stale duplicate at the same version.

Introduced by 03d873557 (2025-07-29, "account for moves in indexing search"). Every scheduled reindex since then has paid it, and it gets worse with every move-heavy edit, which is why the reindex from #1066's migration is so slow now.

## Why move rows exist at all

Only for search deep links. Before 03d873557 search could only link to a version that had fts rows, so a document whose latest version was just a move always showed "Previous Version". Indexing moves also makes a "previous version" link open the last snapshot where the block had that text, in the place it was moved to. fts writes are synchronous by design (search must work as soon as the app opens); only embeddings are async.

## Fix

- `indexChange` collects moved blocks (deduplicated) and resolves them after the ops loop, skipping any block the change replaces or deletes. Each block gets at most one fts row per change.
- `dbFTSGetRawContent` returns only the latest non-empty row: `AND fts.raw_content != '' ORDER BY ftsi.rowid DESC LIMIT 1`. `fts_index` is WITHOUT ROWID with the fts rowid as its primary key, so this is a reverse scan of the existing index with no sort (verified with `EXPLAIN QUERY PLAN` on the vendored 3.45.2 against the prod backup). Semantics are unchanged: the old `onStep` kept the last non-empty row in index order, which is the same row.

On the April prod backup this cuts the lookups from 62,608 to about 1,200.

## Test

`TestIndexChangeIndexesEachBlockOncePerChange`: a change that moves `b1` 500 times and also replaces it, and moves `b2` only. Asserts one fts row per block for the change, `b1` carries the new text (no stale row), `b2` keeps its previous text. Fails on `main` with 502 rows for 2 blocks.

`go test ./backend/blob/... ./backend/api/entities/... ./backend/api/documents/... ./backend/llm/...` and `golangci-lint run --new-from-rev main` pass.

## Not in scope

- The `Commit` duplicate-yield in docmodel (separate PR).
- A composite `fts_index (genesis_blob, block_id)` index would make the remaining lookups exact; unnecessary once the call count drops.
- Peers' `ReconcileBlobs` falling back to the legacy store during reindex adds another 15% of CPU in the same profile; separate issue.

https://claude.ai/code/session_01MR849L65zgoPx52DGBngMD

### DIFF
--- a/backend/blob/blob_change.go
+++ b/backend/blob/blob_change.go
@@ -401,6 +401,22 @@ func indexChange(ictx *indexingCtx, id int64, eb Encoded[*Change]) error {
 
 	extra := changeIndexedAttrs{Actor: uint64(author.ActorID())}
 	opIndex := -1
+
+	// Each block gets at most one fts row per change.
+	//
+	// The CRDT emits one MoveBlocks entry per recorded move, so a block dragged
+	// around during an editing session can appear thousands of times in a single
+	// change. Indexing every entry used to look up the block's previous content
+	// (a scan over every fts row the block already had) and insert yet another
+	// copy, which made the reindex quadratic in the number of moves. Moves are
+	// therefore only collected here and resolved after the loop, and only for
+	// blocks that the change doesn't also replace or delete: those ops write the
+	// block's final content for this version themselves, so a move row would be
+	// a stale duplicate at the same version.
+	ftsWritten := map[string]struct{}{}
+	movedSeen := map[string]struct{}{}
+	var movedBlocks []string
+
 	for op, err := range v.Ops() {
 		opIndex++
 		if err != nil {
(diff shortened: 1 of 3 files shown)


burdiyan:
> The CRDT emits one move entry per recorded move, so a block dragged around in one editing session appears thousands of times in a single change. One prod change lists the same block 2028 times.

I don’t think this is true. At least it wasn’t at some point. We only emit one move per block, so if you drag the block around we only remember the last position. There’s actually tons of complexity to make it work like that.

burdiyan:
Why moves are processed during reindexing at all?

burdiyan:
Isn’t fts indexing asynchronous?

juligasa:
> I don't think this is true. At least it wasn't at some point. We only emit one move per block

Not exactly, Commit yields the same subtree once per old position of its parent (no visited set in the BFS), in prod we have an op moving the same block 1962 times. Will fix that in another PR, I'll correct the description.

juligasa:
> Why moves are processed during reindexing at all?

Only so search links can point to a version where the block was just moved, nothing to do with the doc itself.

juligasa:
> Isn't fts indexing asynchronous?

No, and it shouldn't be, we can't open the app without search. Anyway the issue was one fts read being quadratic (95s of 96s), inserts are 0.7s.

burdiyan:
> Not exactly, Commit yields the same subtree once per old position of its parent (no visited set in the BFS), in prod we have an op moving the same block 1962 times. Will fix that in another PR, I'll correct the description.

Sorry, I don't understand a thing of what you said here 😅. I'll try to poke my AI.

Still, the expectations (the ones I had at least) is that we only emit one move operation per block per change. Full stop. If it's not the case, I'd like to know when it is not the case, and how the hell a Change could've been produced with 1962 move operations on a single block. Separate PR, agree.

> Only so search links can point to a version where the block was just moved, nothing to do with the doc itself.

Why though? Why moving a block should affect the search at all? The content is the same.

> we can't open the app without search

Couldn't disagree more. The app is absolutely useful without search, as long as it can show some data. We can render a progress bar if the user tries to search before FTS is fully ready. This is what macOS does exactly when Spotlight index is being built — you can still use your computer. Also we could even show search results that don't require FTS, like matching document titles, or contacts, and so on.

These are all probably separate concerns, so if whatever you did here fixes the slow reindex, let's merge this.

juligasa:
> Why though? Why moving a block should affect the search at all? The content is the same. 
Not really. Moving a list of cities under a heading that says **Places I like** or moving it below a heading that says **Places I hate** is not the same

On the app without search, you cannot mention, publish, link and more without search. The app would feel kinda broken or we show a read only banner maybe, ... but yes, this can be a follow up PR

Issue #583, opened 2026-05-11
Incorect Citation Count


ericvicenti:
On this URL: https://seedteamtalks.hyper.media/issues/finding-document-after-comment-mention-takes-longer-than-usual-looking-for-this-document-delay/:activity/citations

It says 15 citations, but only one shows up

<img width="956" height="544" alt="Image" src="https://github.com/user-attachments/assets/41f16682-fa0a-475b-be34-1e286fd46a63" />

PR #1088, opened 2026-09-10
Restore historical citations to the activity feed

ion-lion:
## Why now

Issue #583 reports that the citation count is much larger than the citation activity list. The live report still reproduces: `InteractionSummary` counts 24 deduplicated citations, while exhausting the citation `ListEvents` feed returns only the 3 Comment-backed rows.

The two APIs had drifted across Seed's citation representations. `ListCitations` still synthesizes document citations from historical `resource_links.type = 'Ref'` rows sourced by Change blobs and their publishing Ref. The activity route filters only the four modern link types (`comment/Embed`, `doc/Embed`, `doc/Link`, `doc/Button`), so those historical rows never enter the feed. If a Ref-backed mention is returned, the frontend also classifies it as a document update and sends it to `loadRefEvent`, which rejects mention events.

Origin: https://github.com/seed-hypermedia/seed/issues/583

## What changed

- Treat the exact four-type citation request as citation mode and widen only its mention-link predicate to include historical `Ref` links.
- Keep structural Ref blobs out of citation mode, and keep the exact `['Ref']` versions request blob-only.
- Classify `newMention.sourceType = 'Ref'` as a citation while leaving `newBlob.blobType = 'Ref'` as a document update.
- Give citation mode a compound `(time, blob id, link id)` cursor, advanced from the last raw SQL row, so same-timestamp historical links are neither skipped nor repeated after request-local dedup.
- Include source/version/fragment in the mention dedup identity so distinct citations from the same author and timestamp survive.

## Why this approach

Adding `Ref` to the route's shared event-type list is unsafe: the backend applies that list to both structural blob types and resource-link types, which would mix ordinary document updates into Citations. Reindexing or dropping historical citations would also make the activity list disagree with the existing interaction count.

This patch keeps the wire API compatible and narrows the compatibility behavior to the exact existing citation and versions filters. The compound cursor is required because the live historical data contains many citation links with one claimed timestamp; a scalar `timestamp - 1` token permanently skips the remainder.

## Validation

- `CGO_CFLAGS='-O0 -g0' GOMEMLIMIT=320MiB GOMAXPROCS=1 go test -vet=off -p 1 ./backend/api/activity/v1alpha -count=1` — pass.
- Production-shaped backend fixture: Change-sourced Ref citation, publishing Ref, separate target Ref update, four same-timestamp links including one logical duplicate, `PageSize=2`; all pages return three unique citations exactly once, while Versions returns only the target Ref blob.
- Committed frontend classification regression — pass under a bounded Bun harness (1 test, 2 expectations).
- Repository Prettier check for both touched frontend files — pass.
- `git diff --check` — pass.

Immutable validation record: ipfs://bafkreicf2krb7miujswvdqyel2d2fuqb4rszrvdodozefefv2tmaxepx4e

The exact-filter detection is a compatibility shim around an API that conflates structural blob filters with mention-link filters. Replace it when `ListEvents` gains separate first-class blob/link filters, or remove the legacy widening after historical Ref citations are deliberately reindexed to modern link types and parity is verified on existing documents.

### DIFF
--- a/backend/api/activity/v1alpha/activity.go
+++ b/backend/api/activity/v1alpha/activity.go
@@ -82,6 +82,12 @@ type feedCursor struct {
 	key         string
 }
 
+type citationPageToken struct {
+	CursorValue int64 `json:"v"`
+	BlobID      int64 `json:"b"`
+	LinkID      int64 `json:"l"`
+}
+
 var resourcePattern = regexp.MustCompile(`^hm://[a-zA-Z0-9*]+/?[a-zA-Z0-9*-/]*$`)
 
 // NewServer creates a new Server.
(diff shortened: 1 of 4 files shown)


ion-lion:
### Reproduction-first audit

I am closing this PR unmerged. The reporter screenshot shows a mismatch, but I did not independently reproduce the citation-feed behavior on a named baseline before editing, and I did not publish exact-head real-app verification.

PR #1089, opened 2026-09-10
fix(notifications): open document mentions at the source block

### LINKED ISSUE #362: Bug: Clicking on a mention notification redirects to the site homepage instead of the specific mention
https://github.com/user-attachments/assets/3b5761db-2fb6-49ee-b45f-95913b030549

When accessing the notifications section and clicking on a mention notification, the expected behavior is to be redirected to the exact location where the mention occurs. However, the app only redirects to the official homepage of the site where the mention was made, not to the specific mention itself.

**Steps to reproduce:**
1. Open the desktop app and go to the notifications section.
2. Locate a mention notification.
3. Click on the notification.
4. Observe where the app redirects.

**Expected behavior:**
The app should redirect to the specific mention within the document or site, allowing the user to see the context of the mention.

**Actual behavior:**
The app redirects only to the official homepage of the site, not to the mention itself.

**Environment:**

App version: Seed 2026 3.2.
OS: Windows 11
Device: Desktop

ion-lion:
## Why now

Clicking a document-mention notification opens the mentioned document at its top instead of the exact block that contains the mention ([#362](https://github.com/seed-hypermedia/seed/issues/362)). The modern activity event already provides that block ID as `newMention.sourceContext`, but notification construction drops it before inbox persistence. Desktop routing therefore has only the source document UID/path and cannot construct a block-level HM ID.

## What changed

- Retain `newMention.sourceContext` while constructing a modern mention notification.
- Carry it through persisted inbox JSON and the shared notification payload schema.
- Reconstruct the desktop route with `blockRef` when the context is present.
- Preserve historical behavior for omitted, `null`, or empty contexts: those notifications still open the document root.

## Why this approach

This repairs the existing modern `newMention` pipeline at the point where information is lost, rather than inferring an anchor later. It deliberately does not alter the legacy Ref/MentionMap path: that path has a block-ID/account-ID lookup mismatch, and a naive repair could duplicate emails when a publication emits both Ref and modern `newMention` events. Comment mentions also keep their separate `commentId`/`openComment` route.

## Validation

Before rebase, focused tests passed:

- notify: 12/12 (`email-notifier`, payload, persistence)
- desktop: 26/26 (source-context routing and notification behavior)
- includes a direct producer assertion and a production-shaped `citationType: null` route fixture

The commit was rebased without conflicts onto current `main` at `1d35e974`; none of the touched existing production files changed between its original base and current main, and `git diff --check` passes. Initial exact-head CI caught that the new assertion inspected argument 1 of an untyped zero-argument Vitest mock (`TS2493`). The mock now derives its callable signature directly from `Parameters<typeof evaluateMentionEventForNotifications>[2]`, so the assertion remains type-safe as the callback evolves; the complete exact-head frontend check set now passes.

## Follow-up

A real desktop click/scroll smoke test remains useful before merge. Any legacy MentionMap repair should be separate and gated by an exactly-once regression covering paired Ref + `newMention` delivery.

Closes #362

### DIFF
--- a/frontend/apps/desktop/src/pages/__tests__/notification-source-context.test.ts
+++ b/frontend/apps/desktop/src/pages/__tests__/notification-source-context.test.ts
@@ -1,0 +1,23 @@
+import {notificationRouteForPayload} from '@shm/shared/models/notification-helpers'
+import {describe, expect, it} from 'vitest'
+
+describe('document mention notification anchor review', () => {
+  it('routes a document mention to the source block carried by sourceContext', () => {
+    const route = notificationRouteForPayload({
+      feedEventId: 'mention-event-1',
+      eventAtMs: 1_000,
+      reason: 'mention',
+      eventType: 'citation',
+      author: {uid: 'author', name: 'Alice', icon: null},
+      target: {uid: 'source', path: ['doc'], name: 'Source Doc'},
+      commentId: null,
+      sourceId: null,
+      citationType: null,
+      sourceContext: 'block-1',
+    } as any)
+
+    expect(route?.key).toBe('document')
+    if (route?.key !== 'document') throw new Error('expected document route')
+    expect(route.id.blockRef).toBe('block-1')
+  })
+})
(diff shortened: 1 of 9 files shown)


ion-lion:
### Reproduction-first audit

I am closing this PR unmerged. Although the reporter supplied a historical video, I did not independently reproduce the notification-click/scroll failure on a clean packaged-desktop baseline before editing, and there is no exact-head after-video.

PR #1092, opened 2026-09-10
fix(site): let owners hide the home collaborator bar

### LINKED ISSUE #668: Site: allow hiding the "N members collaborating" bar on landing/above-the-fold
On a site's landing page, the "N members collaborating" indicator sits prime above-the-fold, between the top navigation and the actual content.

For sites where collaboration metadata isn't part of the editorial pitch (e.g. a podcast blog whose visitors come for the content, not for the contributor count), this bar competes with the content the visitor came for and pushes the actual article cards below the fold.

Would it be possible to opt out of this bar per document (frontmatter flag) or per site (site setting)? Showing it on some docs (active collaborations, working groups) makes sense; on a landing it doesn't.

Screenshot — bar marked in red on https://lunaticoin.blog:

Image

ion-lion:
## Why now

Issue #668 reports that the always-on “N members collaborating” bar occupies the most valuable above-the-fold space on site landing pages, even when collaboration metadata is not part of the site’s editorial pitch. The bar is rendered only for a space home document, but owners had no persisted control over it.

## What changed

- Add optional home-document metadata `showCollaborators`; omitted remains visible for backward compatibility.
- Add a **Show Collaborators** switch to home-document settings only.
- Honor the flag in both desktop and mobile branches of the shared resource page, so desktop and web render consistently.
- Make document options reachable in the mobile panel, where the menu previously opened discussion content regardless of the `options` route.
- Preserve the setting through typed metadata, Markdown frontmatter, CLI create/update/draft flows, and canonical Markdown export.
- Associate shared switch labels with their controls via the existing `id` prop.

## Why this approach

The bar exists only on the space home page, and Seed already treats that signed home-document metadata as the portable source for site presentation choices such as activity visibility, content width, and header layout. A home-document field therefore behaves as a site-wide setting without adding server-local site configuration that would be lost across gateways. A per-document flag would imply unsupported behavior on pages where the bar is never rendered.

The field is optional and opt-out (`!== false`), preserving every existing site until an owner explicitly changes it.

## Validation

Exact head `2af1121d6c77d671ab6fc4c2abd2cc340c2099a0` on `main` `b378d7e3233ed1a75d23edfc5a786448073ae53d`:

- client Markdown lossless suite: 67/67 passed, including `showCollaborators: false` round-trip;
- client TypeScript check passed;
- CLI TypeScript check passed;
- pinned Prettier check for all ten touched files passed;
- `git diff --check` passed.

- [Exact-head frontend CI](https://github.com/seed-hypermedia/seed/actions/runs/34543570914) passed Typecheck, Lint, web/shared/desktop unit tests, editor E2E, and the aggregate check. Generic integration tests were skipped by workflow design.

The focused local jsdom UI run could not start because an interrupted dependency installation left an incompatible hoisted `lru-cache`; exact-head GitHub CI subsequently executed and passed the committed UI tests.

## Follow-up

If Seed later introduces a distinct signed, portable space-settings resource, migrate this and the existing home presentation fields together rather than creating gateway-local divergence.

Fixes #668.

### DIFF
--- a/frontend/apps/cli/src/commands/document.ts
+++ b/frontend/apps/cli/src/commands/document.ts
@@ -1042,6 +1042,7 @@ const METADATA_KEYS: (keyof HMMetadata)[] = [
   'layout',
   'showOutline',
   'showActivity',
+  'showCollaborators',
   'contentWidth',
   'childrenType',
   'seedExperimentalLogo',
--- a/frontend/packages/client/__tests__/markdown-lossless.test.ts
+++ b/frontend/packages/client/__tests__/markdown-lossless.test.ts
@@ -335,6 +335,7 @@ describe('lossless: metadata', () => {
       seedExperimentalHomeOrder: 'UpdatedFirst',
       showOutline: true,
       showActivity: false,
+      showCollaborators: false,
       contentWidth: 'L',
       childrenType: 'Ordered',
       theme: {headerLayout: 'Center'},
--- a/frontend/packages/client/src/blocks-to-markdown.ts
+++ b/frontend/packages/client/src/blocks-to-markdown.ts
@@ -62,6 +62,7 @@ const FM_KEY_ORDER = [
   'childrenType',
   'showOutline',
   'showActivity',
+  'showCollaborators',
   'seedExperimentalLogo',
   'seedExperimentalHomeOrder',
   'importCategories',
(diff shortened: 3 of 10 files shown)


ion-lion:
### Reproduction-first audit

I am closing this PR unmerged. This mixed an enhancement with a routing defect that I did not first reproduce in the real mobile UI. The static layout screenshot does not demonstrate the options-to-wrong-panel transition.

PR #1095, opened 2026-09-11
Prevent empty comment publication

### LINKED ISSUE #799: Empty comments should not be publishable
Image

ion-lion:
## Why now
Issue #799 reports that the comment UI can publish blank comments. The shared client normalized trailing blank editor paragraphs but then still signed and returned a publishable Comment blob with an empty body. The same gap existed for comment updates, where an empty body is especially ambiguous with the tombstone representation used by deletion.

Closes #799.

## What changed
- Reject new comments when normalization leaves no content, before signing.
- Validate before quote wrapping so an empty quoted reply cannot become a quote-only Embed and bypass the invariant.
- Reject blank comment updates before signing.
- Preserve non-text structured blocks and list containers; the existing trimming classifier only removes empty trailing Paragraph/Heading leaf blocks.

## Why this approach
The invariant lives in `@seed-hypermedia/client`, the shared publication boundary used by desktop, web, mobile, CLI, and pending-intent paths. UI-only button disabling would leave alternate and future callers able to publish the same invalid blob. Validation occurs after editor normalization but before signing, rather than adding a second broad “has content” definition that could reject valid images, embeds, files, lists, or other structured content.

## Validation
- `vitest --run frontend/packages/client/__tests__/comment.test.ts`: 11/11 passed.
- Repository-pinned Prettier 3.1.1: both touched files pass.
- `git diff --check`: passed.

Regression coverage includes zero blocks, whitespace-only paragraphs, quote-only attempts, and blank updates; existing structured-list and quote tests remain green.

No workaround or removal condition. Individual composers may additionally disable submit for UX, but the shared boundary must remain authoritative.

### DIFF
--- a/frontend/packages/client/__tests__/comment.test.ts
+++ b/frontend/packages/client/__tests__/comment.test.ts
@@ -38,6 +38,42 @@ function makeBlocks(text: string): HMBlockNode[] {
 }
 
 describe('createComment', () => {
+  it.each([
+    {name: 'no blocks', content: []},
+    {name: 'a blank paragraph', content: makeBlocks('   ')},
+  ])('rejects $name before signing', async ({content}) => {
+    const signer = makeSigner()
+
+    await expect(
+      createComment(
+        {
+          content,
+          docId: TEST_DOC_ID,
+          docVersion: 'bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi',
+        },
+        signer,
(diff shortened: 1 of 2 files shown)


ion-lion:
### Reproduction-first audit

I am closing this PR unmerged. I did not first demonstrate the real-app empty-comment submission/publication sequence on a clean baseline. The reporter screenshot and unit tests are insufficient for this interaction defect.

PR #1096, opened 2026-09-11
Fix mobile document panel routing

### LINKED ISSUE #809: Document versions and settings on mobile open the comments section.
Repro
1. On iOS Chrome, open: https://seedteamtalks.hyper.media/lobby
2. Try to open either the Document version or settings panel.
3. See the error: 

Image

ion-lion:
## Why now

Issue #809 reports that **Versions** and **Document Settings** both open Discussions on narrow/mobile layouts. The regression came from [`ab84ec405`](https://github.com/seed-hypermedia/seed/commit/ab84ec40582ea3393f5c496575c0ae963e7eb0d4) (2026-02-26): URL routing was expanded so any panel key opens the mobile sheet, while the sheet body remained hard-coded to `DiscussionsPageContent`. Distinct Versions and Options routes therefore reached the right sheet with the wrong content.

Closes #809.

## What changed

- Classify mobile document panels as the specialized Discussions presentation or the shared panel presentation.
- Keep the existing mobile Discussions composer behavior, including delayed keyboard focus.
- Route Versions, Document Options, and other non-comment panel routes through the existing `PanelContentRenderer` already used on desktop.
- Give the filtered Versions activity route its canonical **Versions history** title.

## Why this approach

The routes themselves are already correct and distinct, so changing menu navigation would hide the symptom rather than repair the rendering boundary. Reusing `PanelContentRenderer` avoids a second mobile-only switch and keeps future panel types aligned across responsive layouts. Comments remain specialized because their mobile composer intentionally suppresses focus during the sheet entrance.

## Validation

At exact head `05a906c6eef503c72c6bc7e410484051f727dc88`:

- Before repair, the focused regression failed exactly for Versions (`expected "panel", received "discussions"`), with 48/49 tests passing.
- After repair, `resource-page-common.test.ts` passes 49/49.
- Typecheck, Lint, shared/desktop/web units, editor E2E, and the aggregate frontend check pass in CI.
- Both touched files pass pinned Prettier formatting and `git diff --check`.

No compatibility workaround was added. The comments-specific branch can be removed if `CommentsPanelContent` later gains an explicit mobile focus policy, allowing every mobile route to use the generic renderer without changing keyboard behavior.

### DIFF
--- a/frontend/packages/ui/src/__tests__/resource-page-common.test.ts
+++ b/frontend/packages/ui/src/__tests__/resource-page-common.test.ts
@@ -1,5 +1,6 @@
 import {hmId, type NavRoute} from '@shm/shared'
 import {describe, expect, it, vi} from 'vitest'
+import {createDocumentVersionsPanelRoute} from '../document-versions-panel'
 import {
   getDocumentResourceRouteKey,
   getCommentReplyPanelRoute,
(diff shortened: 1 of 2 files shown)


ion-lion:
### Reproduction-first audit

I am closing this PR unmerged. I did not independently reproduce the mobile panel-routing interaction on a named baseline before editing. A reporter screenshot and component tests do not show the route-changing sequence.

Issue #757, opened 2026-06-10
Comment links are not copied correctly


ericvicenti:
I am seeing many issues with comment links. Range links are not copied correctly, and the comment link often links to the target doc instead of the comment.

On desktop, focus a comment and then try to select the URL in the omnibar. The omnibar is empty when focused so the URL cannot be copied or edited or inspected.

Embedding a comment which has been edited is also broken

ion-lion:
Current-main diagnosis at `2d8bab5bd8b6ff951191ef74b029b0b086847b54`:

Several later repairs now cover the failure modes described here:

- Comment links use the target document's `/:comments/<author>/<tsid>` location, and timestamp navigation was aligned with the copied location in #866 (`82c1904ab`).
- Comment block/range links go through the unified copy-link builder from #500 (`10a3be324`); current tests cover both expanded block links and exact `[start:end]` ranges.
- `<doc>/:comments/<commentId>` embeds resolve to the comment resource after `d624d1705`, while edited versions became routable on web and desktop in #971 (`964234c03`).
- The desktop omnibar now treats `comments` as URL-displayable and its focused state is covered by current component/state tests.

Focused current-main validation is green: shared URL/route suites 252/252, comments UI 5/5, desktop omnibar URL resolution 25/25, URL omnibar 3/3, and desktop omnibar/titlebar state 50/50 (335 total).

I have not closed this yet: I am applying the repository's new UI gate and still need to run the exact click/focus/copy/embed sequence against a real web fixture and packaged Electron build. No product source change will be proposed unless that acceptance reproduces a remaining failure.

ion-lion:
Exact-package desktop acceptance has now crossed the UI gate to **REPRODUCED** at baseline `2d8bab5bd8b6ff951191ef74b029b0b086847b54` (`Seed 2026.9.5-dev.2`).

Observed through Playwright's Electron driver with an isolated production local vault and external exact-package daemon:

1. The fixture document rendered before navigation.
2. URL-mode omnibar accepted the exact edited-comment route.
3. After Enter, repeated renderer snapshots showed the exact `https://dev.hyper.media/hm/.../comment-links/:comments/<author>/<tsid>` URL and the document's Comments panel with count `1`.
4. Neither the original nor edited comment body rendered in any snapshot.

[...]

PR #1102, opened 2026-09-12
Fix focused comments hidden during resource discovery

ion-lion:
## Why now

Issue #757 reports broken comment links and edited-comment embeds. On exact packaged desktop baseline `2d8bab5bd`, the copied-style `/:comments/<author>/<tsid>` route reaches the target document and shows a Comments count of 1, but the focused comment body never renders ([reproduction evidence](https://github.com/seed-hypermedia/seed/issues/757#issuecomment-5641520951)).

The document comment list has already resolved the requested comment, while the independent subscribed Resource fallback can remain in desktop peer discovery. `CommentDiscussions` treated that fallback as authoritative loading state and replaced the valid listed comment with a spinner.

## What changed

- Render the exact focused comment as soon as `ListComments` supplies it, without waiting for the fallback Resource query.
- Accept fallback comment data only when its ID matches the requested comment, preventing React Query's retained previous data from rendering comment A at comment B's route.
- Add focused regressions for both discovery and stale fallback data.

## Why this approach

The route parser, route state, and document comment loader are already correct—the reproduced panel count proves the list data arrived. Changing routing or cache keys would broaden risk without addressing the conflicting loading sources. The repair instead makes the existing list result authoritative and retains Resource discovery only as a fallback for comments absent from the list. Tombstone/deleted-preview behavior is unchanged.

## Validation

- New regression failed on unmodified baseline: 1 failed, 5 passed.
- Focused UI suite after repair: 7/7 passed.
- `git diff --check` passed.
- UI package typecheck was attempted locally but exceeded the 512 MiB executor limit (Node OOM at 256 MiB; process killed at 420 MiB). CI remains required for full type validation.

Packaged acceptance must be rerun unchanged at this PR head before the bug is called verified.

## Follow-up

Rerun the exact packaged desktop acceptance and evaluate web behavior. No workaround or removal condition is introduced; this preserves Resource lookup as the fallback path.

### Packaged verification

`VERIFIED` on the exact PR head `43d0646cc9438b17b6f069ac45fe33680b125960` using the packaged Electron app from [fork-safe artifact run 34665869969](https://github.com/ion-lion/seed/actions/runs/34665869969). The isolated fixture preflight returned HTTP 200 for the home document, target document, and edited comment. The exact `/:comments/<author>/<tsid>` route rendered the edited focused-comment body in the first post-navigation sample, and the titlebar displayed that exact comment route.

Artifact SHA-256: `b952f565ca90908ce00b571c260a333c3e5621aa588d7ecb7d3e12855a0de817`; verification summary SHA-256: `7c896624da511a4c8b5bdb01d43ea3033d1af7e0203528aaec0d626df1e7b827`. A subsequent Playwright `Control+L` probe timed out after the product assertion had passed; this is recorded as harness noise rather than evidence against the focused-comment repair.

### DIFF
--- a/frontend/packages/ui/src/__tests__/comments.test.tsx
+++ b/frontend/packages/ui/src/__tests__/comments.test.tsx
@@ -17,39 +17,39 @@ const {
   useResourceMock,
   onBookmarkToggleMock,
 } = vi.hoisted(() => {
-    const focusedComment = {
-      id: 'alice/comment',
-      version: 'focused-version',
-      author: 'alice',
-      targetAccount: 'alice',
-      targetPath: 'doc',
-      targetVersion: 'document-version',
-      content: [{block: {id: 'text', type: 'Paragraph', text: 'A comment worth saving'}}],
-      createTime: {seconds: 0, nanos: 0},
-      updateTime: {seconds: 0, nanos: 0},
-      visibility: 'PUBLIC',
-    }
-    const parentComment = {...focusedComment, id: 'alice/parent', version: 'parent-version'}
+  const focusedComment = {
+    id: 'alice/comment',
+    version: 'focused-version',
+    author: 'alice',
+    targetAccount: 'alice',
+    targetPath: 'doc',
+    targetVersion: 'document-version',
+    content: [{block: {id: 'text', type: 'Paragraph', text: 'A comment worth saving'}}],
+    createTime: {seconds: 0, nanos: 0},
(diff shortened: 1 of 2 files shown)


ion-lion:
### Reproduction-first audit

I am closing this PR unmerged. I did not publicly capture the complete packaged-desktop comment-link failure before editing. The available prose, JSON, hashes, tests, and package run are not before/after video of the reported interaction.

PR #1103, opened 2026-09-12
Prevent failed discovery from growing scope index

### LINKED ISSUE #957: Bug: arbitrary discovery paths persist empty rbsr_scope rows

ion-lion:
## Why now

Issue #957 shows that each syntactically valid nonexistent discovery path can leave a durable `rbsr_scope` row. The public discovery API accepts remote-only resources by design, so rejecting unknown IRIs before synchronization would break legitimate discovery. With scope eviction currently disabled, arbitrary unique paths can otherwise grow the derived index without bound.

Closes #957.

## What changed

After cold materialization, the maintained index now discards a scope only when it has neither eligible `rbsr_item` rows nor a locally known resource. Known-but-empty resources remain indexed so incremental maintenance can populate them when blobs arrive.

The RBSR tables are derived, so a migration rebuilds them and clears already-accumulated empty rows. Scope IDs now use SQLite `AUTOINCREMENT`; this prevents a deleted ID from being reassigned while an in-flight reader or shadow verifier still carries that numeric identity.

## Why this approach

- An API-level existence check was rejected because remote-only resources must remain discoverable.
- Deleting every empty scope was rejected because known resources can legitimately have no currently eligible/downloaded blobs and still need incremental maintenance.
- Plain row deletion with `INTEGER PRIMARY KEY` was rejected because SQLite may reuse the deleted maximum ID, racing code that carries scope IDs across transactions.
- The migration rebuild is appropriate because both RBSR tables are derived and lazily rematerialized; it also removes historical pollution without a separate cleanup scan.

## Validation

- Regression on unmodified main `4b356bfb29a801a6ad310e16942e3f1dc72a66fd`: `TestLoadIndexedScopes_DropsEmptyScopes` fails because one empty row persists.
- Focused fixed tests pass, including repeated missing-path retries, non-reused IDs, known-empty resource retention, and legacy-store parity.
- `go test -vet=off -p 1 ./backend/hmnet/syncing -count=1` passes.
- `go test -vet=off -p 1 ./backend/storage -count=1` passes, including snapshot migration parity.
- `git diff --check` passes.

`./dev gen //backend/...` could not run in this executor because the checkout lacks the repository's required direnv/Please environment. Draft CI remains the authoritative generated-code and complete backend check.

## Follow-up

Request rate limiting can independently bound CPU/network work before materialization. It is not required to stop durable empty-row growth and should be evaluated separately rather than coupling API admission policy to this derived-index fix.

### DIFF
--- a/backend/hmnet/syncing/rbsr_index.go
+++ b/backend/hmnet/syncing/rbsr_index.go
@@ -233,6 +233,16 @@ func loadIndexedScopes(ctx context.Context, db *sqlitex.Pool, dkeys colx.HashSet
 					if err := materializeScope(conn, id, dkey); err != nil {
 						return err
 					}
+					shouldKeep, err := scopeHasItemsOrResource(conn, id, dkey)
+					if err != nil {
+						return err
+					}
+					if !shouldKeep {
+						if err := sqlitex.Exec(conn, qDeleteScope(), nil, id); err != nil {
+							return err
+						}
+						continue
+					}
 				}
 				scopeIDs = append(scopeIDs, id)
 			}
@@ -319,8 +329,22 @@ const qMaterializeReplace = `
 
 var qMaterializeClear = dqb.Str(`DELETE FROM rbsr_item WHERE scope = :scope;`)
 
+var qScopeHasItemsOrResource = dqb.Str(`SELECT
+	EXISTS(SELECT 1 FROM rbsr_item WHERE scope = :scope)
+	OR EXISTS(SELECT 1 FROM resources WHERE iri = :iri);`)
+
+var qDeleteScope = dqb.Str(`DELETE FROM rbsr_scope WHERE id = :scope;`)
+
 var qMarkMaterialized = dqb.Str(`
 	UPDATE rbsr_scope SET materialized = 1, last_access = :now WHERE id = :scope;`)
+
+func scopeHasItemsOrResource(conn *sqlite.Conn, scopeID int64, dkey DiscoveryKey) (keep bool, err error) {
+	err = sqlitex.Exec(conn, qScopeHasItemsOrResource(), func(stmt *sqlite.Stmt) error {
+		keep = stmt.ColumnInt64(0) != 0
+		return nil
+	}, scopeID, string(dkey.IRI))
+	return keep, err
+}
 
 // materializeScope (re)builds a scope's persisted blob set from collectBlobs —
 // the one place the expensive closure runs. Idempotent: it clears and refills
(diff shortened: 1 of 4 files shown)


ion-lion:
### Reproduction-first audit

I am closing this PR unmerged. I did not first publish a real-daemon/API and database transcript showing the reported failed-discovery row growth. Unit/integration tests alone do not prove the reported behavior.

PR #1104, opened 2026-09-12
Fix duplicate comment mention notifications

### LINKED ISSUE #703: Mention creates duplicate notif
Comment on a document and mention an author. The author will get two notifs but should only get one

ion-lion:
## Why now

Issue #703, reported on 2026-06-01, identifies a user-visible delivery gap rather than a new regression: a comment that mentions the document author can generate two notifications. Current activity construction intentionally emits both the comment blob and a companion `newMention` event. The notifier already used same-batch source-CID detection to avoid extracting the body mention twice, but normal comment precedence still classified the blob as a discussion/reply for the owner or parent author, while the companion event separately classified as a mention.

## What changed

- Map dedicated `comment/*` mention events by source comment CID and recipient account.
- Defer the comment-blob notification only when that exact recipient is both mentioned in the parsed comment body and covered by a companion mention event.
- Keep fallback body classification for recipients/events without a companion mention, and preserve document citations (including account-home document citations).

## Why this approach

Deduplicating only by feed event ID cannot work because the blob and mention are intentionally distinct events, and it would still send duplicate immediate emails. Globally suppressing comment-blob mentions would lose compatibility when a companion event is absent. Recipient-specific precedence preserves those fallbacks while making the dedicated event authoritative for the overlapping logical mention.

## Validation

- Baseline focused regression: 1 failed / 12 passed (`discussion` returned where the companion mention should own delivery).
- Exact-head focused tests: 27/27 passed across notifier event mapping/precedence and shared comment classification.
- Exact-head CI at `e37299ec2`: Typecheck, Lint, web/shared/desktop units, editor E2E, and aggregate checks passed: https://github.com/seed-hypermedia/seed/actions/runs/34703819761
- Prettier and `git diff --check` pass.

The source-CID fallback can be simplified only if the activity API guarantees and tests atomic delivery of each comment blob with all companion mention events. Until then, retain body-based fallback for uncovered recipients.

Closes #703

### DIFF
--- a/frontend/apps/notify/app/email-notifier.test.ts
+++ b/frontend/apps/notify/app/email-notifier.test.ts
@@ -4,7 +4,9 @@ import {Code, ConnectError} from '@connectrpc/connect'
 import type {HMBlockNode} from '@seed-hypermedia/client/hm-types'
 import type {Event} from '@shm/shared'
 import {
+  getDedicatedCommentMentionTargets,
   getEventId,
+  hasDedicatedCommentMentionEvent,
   getMentionsOfDocument,
   isNotificationEventTooOld,
   isTransientGrpcUnavailableError,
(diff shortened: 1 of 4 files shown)


ion-lion:
### Reproduction-first audit

I am closing this PR unmerged. I used a focused test as mechanism evidence without first publicly reproducing the real mention/comment sequence and duplicate user-visible delivery. Exact-head CI is not post-fix user-level proof.

Issue #598, opened 2026-05-13
Desktop: Back button after publishing new doc says "not found"


ericvicenti:
The back button after you publish a new doc will take you to a "not found" page as it goes back to the draft's temporary pathname, which is no longer existing as a document or a draft

This should be fixable by doing a replace to the new path after you publish, instead of doing a push

ion-lion:
Verified on exact packaged Electron builds with a byte-identical route-aware acceptance.

- Current main `4b356bfb2`: Back returned to the deleted temporary `/-<draft-id>` route (**reproduced**).
- Fix `e93b9a563`: Back returned to the pre-creation home document (**verified**).
- PR: https://github.com/seed-hypermedia/seed/pull/1109
- Durable evidence: `hm://z6MkpVa5nMUR5ZaUEyV1SE48KTNbwTuHRd83RwLgMKc4nGU3/ion/progress/desktop-publish-back-navigation-598`
- Content-addressed bundle: `ipfs://bafkreialra7asrm2tkj72cvzcmam7vycu7kjtjynibtm5j36qoo4t6xsby` (SHA-256 `0b883e09459a9a93fd0ab91300cfd702a7d499a70d4066cea77e839dc9faf20e`)

Issue #438, opened 2026-04-07
"View Deleted Content" feature


ericvicenti:
Do we support the ability to see what has been deleted? I just discovered this feature in the app but its totally broken!

We should probably delete this code because it is totally not working 😂

<img width="1812" height="1175" alt="Image" src="https://github.com/user-attachments/assets/9b1e9f2e-33e5-43fc-826e-331a09a0b0fa" />

horacioh:
## Germinator triage

**Root cause hypothesis:** The `ListDeletedEntities` and `UndeleteEntity` gRPC handlers are fully commented out in the backend, so the server falls through to the `UnimplementedEntitiesServer` stubs which return `UNIMPLEMENTED` to every call — the entire feature runs on dead gRPC endpoints.

**Likely files:**
- `backend/api/entities/v1alpha/entities.go:1747` — `ListDeletedEntities` handler is fully commented out; the `Server` struct embeds `UnimplementedEntitiesServer` so calls hit the stub at `genproto/entities/v1alpha/entities_grpc.pb.go:197` which returns `codes.Unimplemented`.
- `backend/api/entities/v1alpha/entities.go:1735` — `UndeleteEntity` handler is also fully commented out, so the undelete button on the list page also fails identically.
- `frontend/apps/desktop/src/pages/deleted-content.tsx:12` — `useDeletedContent()` calls `listDeletedEntities({})` which will receive a gRPC UNIMPLEMENTED error; the page has no error state and renders `deleted.data || []` as an empty list.

**Reproduction steps (expanded):**
1. Open the desktop app and navigate to **View → Review Deleted Content** (or use the menu item registered in `frontend/apps/desktop/src/app-menu.ts:171`).
2. Observe that the list never populates — the gRPC call returns `UNIMPLEMENTED` and `useQuery` surfaces an error but the component only renders `deleted.data || []`, resulting in a silently empty list.
3. If any item somehow appeared, clicking the undelete button would also fail because `UndeleteEntity` is similarly commented out.

**Plausible fix sketch:**
Uncomment the two handlers in `backend/api/entities/v1alpha/entities.go`:
- `ListDeletedEntities` (lines 1747–1770) iterates `hypersql.EntitiesListRemovedRecords` and maps the rows into `DeletedEntity` structs — the SQL query layer appears ready, so just removing the `//` comment prefixes should restore it.
[...]

PR #1107, opened 2026-09-13
Add reproducible exact-source desktop artifacts

ion-lion:
## Why now
Repeated frontend bug gates need Linux Electron packages built from an exact source SHA. The constrained local executor cannot package them reliably, so recent acceptance work used one-off workflow mutations. During #438, a brittle multiline `printf` mutation produced an invalid package workflow and delayed verification.

## What changed
Adds a shared agent skill whose single `acquire` command launches, binds, polls, downloads, hashes, and verifies an exact-source package on Ion's fork. The evidence JSON records the run, workflow commit, source SHA, artifact identity, SHA-256, and verified embedded source manifest.

## Why this approach
The workflow commit is verified byte-for-byte against the requested source and issue before a run is accepted. Checkout credentials are not persisted, the PAT is removed before GitHub's cross-origin artifact redirect, and both a supported non-empty installer and exact source manifest are required. This replaces issue-specific YAML editing without requiring a persistent self-hosted packager.

## Validation
- `python .agents/skills/exact-desktop-artifact/scripts/test_exact_desktop_artifact.py` — 7/7 pass
- Python bytecode compilation passes
- Credentialed read-only fork/main ref inspection passes

A real package launch is intentionally deferred until the next UI gate needs one, avoiding CI spend solely to test the harness.

Exercise `acquire` unchanged on the next exact-source desktop gate, then incorporate any operational findings. The fork-specific launcher can be generalized or removed if Seed gains a maintained workflow-dispatch package builder that accepts and attests an arbitrary source SHA.

### DIFF
--- a/.agents/skills/exact-desktop-artifact/SKILL.md
+++ b/.agents/skills/exact-desktop-artifact/SKILL.md
@@ -1,0 +1,25 @@
+---
+name: exact-desktop-artifact
+description: Build and verify an exact-source Linux desktop package on Ion's GitHub fork
+---
+
+# Exact desktop artifact
+
+Use this when UI acceptance requires packaged Electron at an exact 40-character source SHA and local packaging is impractical.
+
+```sh
+python .agents/skills/exact-desktop-artifact/scripts/exact-desktop-artifact.py acquire \
+  SOURCE_SHA ISSUE_NUMBER /safe/output/artifact.zip
+```
+
+The command creates or updates a dedicated workflow on `ion-lion/seed`, binds the run to the workflow commit, waits for success, downloads the uniquely named artifact, checks GitHub's digest when present, and verifies the embedded `ION-SOURCE-SHA.txt` plus a non-empty Linux installer. It prints a JSON evidence record with run URL, source SHA, artifact SHA-256, and local path.
+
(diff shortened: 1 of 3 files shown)


ion-lion:
Real issue-driven operational proof is now complete for #598.

The first genuine acquisition exposed two blockers that offline mocks missed:

1. GitHub's Contents API returns base64 with line wrapping; strict decoding now removes ASCII whitespace first while still rejecting malformed content.
2. The 627 MB package artifact exceeded this executor's 512 MiB memory when buffered. Downloads now stream from the credential-free approved redirect into a bounded temporary file while hashing, verify the ZIP and embedded source manifest on disk, then atomically replace the destination.

Validation:
- harness tests: 9/9 pass
- Python compilation and `git diff --check` pass
- exact-main source: `4b356bfb29a801a6ad310e16942e3f1dc72a66fd`
- run: https://github.com/ion-lion/seed/actions/runs/34736620180
- artifact: 627,432,607 bytes, SHA-256 `c2b137728a98704c8feaad4339073127d5944b5dd1c97779608e441395476d3a`
- embedded source manifest: verified

The same repaired harness has launched the exact-fix #598 package run: https://github.com/ion-lion/seed/actions/runs/34737741282.

PR #1112, opened 2026-09-14
fix: disambiguate colliding metadata column labels

### LINKED ISSUE #1110: Status column can be enabled twice
I think one of these is `Status` and the other one is `status`

Image

What is the best way to clean this up?

ion-lion:
## Why now

Issue #1110 reports two independently enableable table columns that both render as `Status`. The metadata identities are actually `Status` and `status`: Seed intentionally preserves case-sensitive custom metadata keys, but the query-table presentation formatter capitalizes both labels and makes them indistinguishable.

## What changed

- Detect collisions after formatting discovered metadata labels.
- Keep friendly formatting for unique keys.
- Fall back to each exact raw key only for colliding labels, preserving both identities and values without rewriting signed document metadata.
- Add a model regression for `Status` and `status`.

## Why this approach

Lowercasing, merging, or migrating arbitrary metadata keys would be destructive because backend attribute identity is case-sensitive. Disambiguating only the display label fixes the reported ambiguity while retaining stable `metadata:<exact-key>` IDs, persisted visibility settings, and sort/value lookup behavior.

## Validation

- Exact-main source `452cb932c478d6dd7f6c7c3f7dc0651014c33319` reproduced two distinct IDs with duplicate display labels: `Status`, `Status`.
- Exact-head source harness returns distinct labels: `Status`, `status`.
- `git diff --check` passes.
- Full workspace Vitest is pending: dependency installation exceeded this runner's memory limit, so this PR remains draft.

## Follow-up

Run the committed focused Vitest regression and exact UI acceptance in a provisioned frontend runner, then mark ready only if both pass unchanged.

Fixes #1110


ericvicenti:
this fix would technically help but I think the real issue is that we need schemas + UI that helps ensure every task uses the same fields in the same way

ion-lion:
Agreed. This patch only disambiguates the inconsistency after it has already been created; it does not help a task collection converge on one field identity.

I’m closing this draft and leaving #1110 open. The next implementation should start from an explicit task schema (canonical field IDs, types, and options) and have task authoring choose from that schema, while treating existing arbitrary signed metadata as legacy/imported data rather than silently merging it.

PR #1119, opened 2026-09-14
fix(blob): index object-valued document metadata

### LINKED ISSUE #1117: fix(documents): object-valued metadata is silently dropped from the document index
## Summary

Document attributes whose value is a nested object are silently excluded from the `document_attributes` index. As a result, `QueryDocuments` / `ListDocuments` / `GetDocumentInfo` never return object-typed metadata keys, even though `GetDocument` returns them correctly. This breaks the schema inference added in #1116: `inferDocumentSchema` never sees nested attributes, so new collection children only receive primitive defaults.

## Reproduction

1. Publish two children of a collection, each with a nested custom key, e.g. `customer: {name, score, active, address: {city}}`.
2. `GetDocument` on each child returns `customer` intact.
3. `QueryDocuments` for the collection's children returns `metadata` with no `customer` key at all — not even flattened dotted keys.
4. New → Document in the collection creates a draft whose metadata has the inferred primitives (`""`, `0`, `true`) but no `customer`.

## Expected

Object-valued attributes are indexed (e.g. flattened into dotted leaf keys like `customer.name`), so list/query metadata matches `GetDocument` and downstream features can rely on nested shapes.

## Actual

`documentAttributeValue` (`backend/blob/blob_ref.go` ~line 925) only accepts null/string/bool/int; `setAttributeOrdered` (~line 997) swallows the error and drops the attribute. The index row never exists, so `documentInfoFromRow` → `attrs.PublicMap()` has nothing to rehydrate.

Note that `PublicMap` already unfolds dotted attribute keys into nested maps, so flattening object values into `parent.child` leaf attributes at index time would both fix the listing and round-trip back to the nested shape.

Verified on `8b138e36e` against a live daemon; the `document-creation-schema` unit tests pass, so the gap is in the indexing path, not the schema logic.

ion-lion:
## Why now

Issue #1117 reports that object-valued document metadata survives `GetDocument` but disappears from `QueryDocuments`, `ListDocuments`, and `GetDocumentInfo`. The blob index accepted only scalar values, so the structured value never reached `document_attributes`; that also prevents schema inference from observing nested metadata.

## What changed

- Flatten object-valued attributes into dotted scalar leaf paths before persistence, which `DocIndexedAttrs.PublicMap` already rehydrates into nested maps.
- Emit a same-order parent tombstone before an object's leaves. This preserves replacement semantics: fields omitted by a newer object do not survive as stale indexed children, while the same-order descendants replace the temporary tombstone.
- Continue ignoring arrays and other unsupported structured values. Scalar FTS and IPFS-link indexing behavior is intentionally unchanged.
- Replace the old “ignore structured attributes” regression with a two-change round-trip test covering nested strings, integers, booleans, stale-leaf removal, and unsupported arrays.

## Why this approach

Adding an object storage kind would require a schema/read-path migration and duplicate the existing dotted-path hydration mechanism. Flattening at ingestion uses the representation the index and public API already understand. The parent tombstone is necessary because leaf-only flattening would merge replacements and retain omitted fields.

## Validation

- `git diff --check` passes.
- Focused local Go test compilation was attempted with Go 1.26.2, but the 512 MiB executor killed `gcc` while compiling `backend/util/sqlite`; exact-head CI provides the full-memory validation.
- Exact-head `lint-go` and `test-go` both pass at `0ab5ef3eb55e1ad9e173777567f7bb7cc2d03ae8`.

## Follow-up

Empty objects cannot be distinguished from removal in the scalar dotted-path index. If clients require empty-object round-tripping, that should add an explicit object storage kind and migration rather than overload a scalar sentinel.

Closes #1117


ion-lion:
### Reproduction-first audit

I am closing this PR unmerged. Object-valued document metadata was clarified as unsupported input, so this is not an established defect. It should not be merged under a bug-fix framing; any future work must begin as an explicit API proposal.

PR #1125, opened 2026-09-15
fix(daemon): comments belong to their location, not their genesis

ericvicenti:
> **Stack:** this is PR 1 of 2. #1126 (SDK genesis fix) is based on this branch. To test both together, check out `fix/sdk-document-genesis`. Merge this first, then #1126.

## Symptom

On https://ht26.hyper.media/pro/3800935.3830850 and every other paper under `/pro`, the same three comments show on every document. Same on the source site the papers were republished from.

## What's actually wrong with the data

All 60 papers on the source site share one genesis (`bafyreigcjeqiv6luf2la46i…`) and report a 1970 creation date, and every ht26 republish carries that same genesis. A genesis is a document's first Change, and the SDK's `createGenesisChange` builds it from the signer alone with a zero timestamp, which is deliberate for the account **home** document (any device bootstraps the same one) but was reused for ordinary documents by the agents API's `writeDocumentCreate` and the CLI test helper. Every document one account creates through those paths gets the home document's genesis. Fixing that SDK misuse is a separate PR; this PR makes the daemon correct on the data as it is.

## Why comments merged

The daemon identified a document for comments by its genesis (`comment_live.genesis`, `document_comment_stats` keyed by genesis). That is what let a moved document keep its thread, but sixty documents on one genesis are one document under that rule, and a republish Ref borrows its target's genesis, so a republish shared the original's thread too.

## New rule

Comment identity is the **location**, with moves carrying comments and republishes not:

- A document owns the comments written at its path, plus the comments written at every path that was **moved** into it (transitively, bounded at 16 hops).
- The walk first follows move redirects **forward** from the queried path to the document's current location, so querying a path the document moved away from still returns its whole thread, then gathers **backward**.
- A **republish** is not a move (its redirect Ref is alive; a move's is a tombstone). It starts its own thread.
- A redirect to an **unrelated** document does not carry comments. The daemon stamps every redirect Ref with its target's genesis, so the hop can't tell; instead readers keep only comments written against the root's genesis, which a moved document keeps and an unrelated one lacks. This preserves the guard the previous genesis-based design was introduced for (the 293-comment `/tech-talks` → `/tech` case in the code comments).

`ListComments` and the listing's comment activity (`comment_count`, latest comment) use one shared walk (`commentLineageSQL`). The listing evaluates it in the outer layer of `wrapDocumentsQuery`, so only the final page pays. `document_comment_stats` stays genesis-keyed and still feeds `activity_time` ordering; that is the one place shared-genesis documents still influence each other (sort position), noted as a follow-up.

## Verification

- Go: `TestCommentCount_DocumentsSharingAGenesisKeepSeparateThreads` builds the bad-data shape (two papers sharing a genesis with different content, plus a republish), comments on each, and requires each path to list and count only its own comment. `TestCommentCount_RepublishRedirect` now encodes the republish rule. `TestCommentCount_AfterDocumentMove` (multi-hop moves, query by old path, delete decrements) and `TestCommentActivityFollowsTheDocument` (unrelated redirect must not transfer) pass unchanged. Full `documents/v3alpha` package green; lint on new code clean.
- End to end: the CLI fixture suite gains "documents that share a genesis keep separate comment threads". Its documents come from the test helper that still has the SDK misuse, so they share a genesis for real; a comment on A must not appear on B. Fails against the old daemon (B listed A's comment, relabeled as targeting B), passes against the new.
- Cost, best of 20 on a test daemon: ListDirectory of 60 rows 1.37 ms → 3.20 ms; ListComments 40 µs → 66 µs.

## Not in this PR

- The SDK fix (explicit home-genesis vs document-genesis helpers; switch `writeDocumentCreate` and the CLI test helper). Existing collided documents keep their shared genesis either way; this PR makes them behave correctly regardless.
- `activity_time` ordering still keyed by genesis.

https://claude.ai/code/session_01QiGioTPhh6qUQedRpo5m4q


burdiyan:
Why all 60 papers share the same genesis? Are they fork of the same document? Does it make sense at all for them sharing the same genesis?

We explicitly made comments belong to Genesis to properly list comments from all the forks and accurately aggregate comment counts. I would like to keep it this way, we’ve been talking about doing this forever.

If a has all of the documents sharing the same Genesis for no reason I think they are “holding it wrong”.

juligasa:
I agree with @burdiyan, lets keep comments by genesis and fix the data instead (merge #1126, recreate the 60 papers). If we want I can add a small guard at publish time to reject refs on the zero-ts home genesis with a non empty path, so this cannot happen again.

ericvicenti:
ok, sounds good. Will fix the root cause where content was getting created by agents with root genesis #1126

PR #1127, opened 2026-09-16
fix(web): show the comment box on republished documents

ericvicenti:
## Problem

On the web, a republished document has no comment composer at all: no editor, no loading state, no error. Repro: https://hyper.media/hm/z6MkiAKDcRSzQ4zPZfnJcS5HYx5MwgN6MU9foHihJGrhqNBj/for-julio/:comments (a sibling non-republished doc like `/notes/:comments` shows the box).

## Cause

- `queryResource` follows the republish redirect and returns the target document under the republish's **own** id, whose `version` is null. The web loader keeps that id unpinned on purpose (pinning the target's version onto the republish address asks the daemon for a version that does not exist there).
- `WebCommenting` read `docId.version` and returned `null` when it was falsy.

Confirmed by reading the mounted editor's props in the browser: `docId.version === null` on the republish, a CID on the sibling doc, same origin and config.

## Fix

Resolve the version from the fetched resource first and fall back to the id (`resolveCommentTargetVersion`, small helper with unit tests). This mirrors the desktop comment editor, which already uses `targetDoc?.version || docId.version`. While the resource is still loading, the existing "Loading…" placeholder is shown instead of nothing.

The comment payload is unchanged in shape: it targets the republish's address with the target document's version, exactly as desktop does today.

## Verification

- Local dev server + fresh local daemon (discovered the republish and its target from the network): the composer renders on `/for-julio/:comments`, clicking it opens the editor. `/notes/:comments` unchanged.
- `pnpm typecheck` and `pnpm test` in `frontend/apps/web` pass (251 + 3 new tests).
- Not exercised: actually publishing a comment from a republish page (would post to a real document).


ericvicenti:
This is not my favorite solution but it makes sense

PR #1128, opened 2026-09-16
fix: resolve version-pinned lookups at a republish address (embed of a republish failed to load)

ericvicenti:
## Problem

An embed of a republished document rendered "Could not load embed", while an embed of the original worked. The pinned republish URL (`/hm/<site>/doc?v=...`) on the web also 404ed.

The republish's web page advertises the target's version under the republish's own address, so a pasted or shared link becomes `hm://<republish-site>/doc?v=<target version>`. That is what the embed block stores.

For that lookup the daemon reported NotFound. An unversioned `GetDocument` goes through `ResolveLatest`, which sees the redirect Ref at the path and returns the redirect error. A versioned `GetDocument` skips that and replays the requested change CIDs at the path via `IterChanges`; a redirect path has no change DAG, so it found nothing and returned NotFound without ever consulting the redirect. Same path, two different answers depending only on whether `?v=` was present.

## Fix

- **daemon**: `GetDocument` and `GetResource` route a NotFound from a versioned load through `redirectOrError`, which runs the `ResolveLatest` lookup the versioned path skipped and returns the existing redirect error if the path holds a redirect. Every other error and every unversioned load is untouched. Cost: two indexed reads, only on the versioned-NotFound path.
- **shared**: `queryResource` and `createResourceResolver` carry the version across a republish hop (`republishVersionCarry` in `redirects.ts`), since the version belongs to the target. Move redirects keep dropping the version, as before.
- **web**: the SSR loader follows a version-pinned republish route the same way, which fixes the 404.

## Tests

- Go: `TestRepublishDocument` now asserts that a versioned `GetDocument` and `GetResource` at the republish path report the republish redirect, and that a versioned lookup at a plain missing path is still NotFound. The new assertions fail without the daemon change.
- vitest: new cases for `queryResource` (carry on republish), `createResourceResolver` (carry on republish, no carry on move), and the web loader (pinned republish route resolves to that version of the target).

Full `documents/v3alpha` Go package, the whole `@shm/shared` suite, web loader tests, and typecheck for shared/web/desktop pass.

Repro: https://hyper.media/hm/z6MkiAKDcRSzQ4zPZfnJcS5HYx5MwgN6MU9foHihJGrhqNBj embeds `hm://z6MkiAKDcRSzQ4zPZfnJcS5HYx5MwgN6MU9foHihJGrhqNBj/for-julio?v=bafy2bzacebv6...`, a republish of https://beefsteak.hyper.media/for-julio. Needs a daemon deploy to take effect on production sites.


ericvicenti:
@burdiyan @juligasa the backend change here is meant to solve the situation when a version is provided and you are looking up a republished doc. instead of thowing a 'NotFound' error, it should throw a redirect error.

burdiyan:
I think the expected UX would indeed be resolving the document properly, and preserve the `v`. My concern though is that this patch makes the daemon follow redirects now, which used to be the concern of the client. There was a reason why we decided to resolve redirects on the client and not on the server, but I forget what those reasons were.

For better coherence I think we should make `GetDocument` and `GetResource` return the redirect error, like we do for non-versioned requests. And then the client shall resolve the redirect target, preserving the `v`. Resolving the redirect target on the backend only for this specific edge case I think would be an inconsistency.

ericvicenti:
Glad we agree on the approach, and that is what this does: the daemon does not follow the redirect. `redirectOrError` returns the same FailedPrecondition redirect error the unversioned path produces, and nothing else:
https://github.com/seed-hypermedia/seed/blob/a97eca2d40f74182eae9bdf37238447e6bab7999/backend/api/documents/v3alpha/documents.go#L923-L933

The test asserts a FailedPrecondition with `RedirectErrorDetails`, not a document:
https://github.com/seed-hypermedia/seed/blob/a97eca2d40f74182eae9bdf37238447e6bab7999/backend/api/documents/v3alpha/documents_test.go#L2925-L2933

Following the hop and preserving `v` happens on the client:
https://github.com/seed-hypermedia/seed/blob/a97eca2d40f74182eae9bdf37238447e6bab7999/frontend/packages/shared/src/redirects.ts#L20-L26

burdiyan:
Oh, sorry. I misunderstood the PR description. LGTM then.

PR #1130, opened 2026-09-17
fix(agents,ui): agent-published docs show up on the desktop right away

juligasa:
When an agent publishes from the hosted server, the doc lands on hyper.media and the desktop has to pull it. Two gaps:

- If the node already had an older version, the live "latest" discovery was happy with the stale copy. Now the write's `?v=` is pinned on a bounded probe. Measured on a mainnet desktop: 1.1s vs "whenever the next re-run happens".
- If the session wasn't on screen (background, trigger, or you just navigated away), nothing asked at all. The server now announces write results on the account hint with the published refs, and the title-bar socket runs the same bounded probe. No live subscriptions from that socket, so nothing accumulates.

Needs the hosted agent server redeployed for the second part; old servers/clients are unaffected either way.

Tests in `api-service.test.ts` and `agents-models.test.tsx`.


ericvicenti:
Reviewed this. The diagnosis is right and I confirmed the daemon side: a task that names a version only settles when the result matches (`scheduler.go:132`), it re-runs immediately while it does not (`scheduler.go:523`), and an already-local version returns straight from `GetResource` (`discovery.go:161`). So replaying old write results after a reconnect is cheap, and deletes return no version so they fall back to a plain discover.

Three small fixes I am pushing on top:

- A failed `discoverEntity` call ended the probe (`catch` calls `onSettled`). One busy daemon or timeout loses the published version for the rest of the window. It now retries on the same cadence until the deadline.
- A comment write result references the comment at its own version, which canonicalizes to the target document, so the session path pinned the document to a version it can never report and ran the full 90s. The account-hint path already gets this right, the session path now follows the same rule.
- The session's reference sync and the account hint for the same write each ran their own probe on separate 5s cadences. Both now hold the shared probe registry. That also covers the case where the document is already subscribed (a comment made it recursive, or the earlier reference had no version), where the version was previously skipped entirely and the live subscription only ever asks for "latest".

## Cleanup worth doing

After this lands there are four overlapping paths that get agent content onto the local node: the session reference sync, the new account hint, `useAgentAccountsSync`, and the one-shot `syncAgentAccountToLocalNode`. Peering is doubled too, `useConnectLocalNodeToAgentHmServer` inside every socket hook and `useConnectLocalNodeToAgentHmServers` inside `useAgentAccountsSync`.

1. **Make the account hint the only path for published content.** The title bar keeps one account socket per server for the life of the window (`AgentAccountLiveUpdates`), so the hint always lands. That makes `subscribeToAgentReferences` and its live subscriptions redundant, along with the hm:// scan of assistant prose, which only ever covered links to content the agent did not write, and clicking one subscribes on its own. The cost is that an agent server that has not been redeployed stops syncing even for an open session, so this probably wants to be gated on the server's protocol version.
2. **Handle `references` in one place.** Every `account/`-keyed `useAgentWebSocketSubscription` runs the probe today: title bar, agents list, server page, and the second account socket on the session page. That is the only reason the shared registry needs holder sets and ref-counting. If only `AgentAccountLiveUpdates` handles it, there is one owner per server per window and the probes can be plain fire-and-cancel-on-unmount.
3. **Reuse the retry loop that already exists.** `discoverDocument` in `app-sync.ts:734` already asks for a version and retries until it matches. `pinAgentReferenceVersion` rebuilds that in the renderer. Exposing it as a `sync.discover` procedure would delete the hand-written loop. The same place is where the root cause lives: `runDiscovery` passes `version: undefined` even when `sub.id.version` is set, so every `?v=` link in the app has this same stale-copy problem, not just agent content.
4. **Shrink `useAgentAccountsSync`.** Drop `useConnectLocalNodeToAgentHmServers`, since the title-bar sockets already keep every server peered. Once writes arrive on hints, its remaining job is keeping signer and owner profiles available for @mentions, which does not need recursive `/**` subscriptions on whole accounts. And mount it once in the title bar instead of in four pages, which is what its own doc comment promises anyway.

1 and 2 could go in this PR. 3 and 4 are a follow-up, and 3 touches `app-sync.ts` enough to deserve its own review.

juligasa:
Sorry for the confusion the other day. I Could not reproduce the minutes delay on mainnet: subscriptions land in ~10s foreground / under a minute background, discover takes ~1s, the only gap is the desktop not asking for off-screen sessions and that is what this fixes.

ericvicenti:
Ok but the bug I experienced was an on-screen session who linked to a document(or comment, I forget). And the content did not sync.

For off-screen sessions, I'm not sure if its the desired behavior to immediately sync, but it should be fine until people start experiencing bloat.

So, please merge it!

PR #1126, opened 2026-09-15
fix(client,agents,cli): give every new document its own genesis, not the home document's

ericvicenti:
> **Standalone.** Rebased onto main after #1125 was closed; the fixture case that depended on #1125's daemon change was dropped. Only the SDK, agents, CLI test helper, tests and the CI matrix entry remain.

## The bug

`createGenesisChange(signer)` in `@seed-hypermedia/client` builds an empty first Change from the signer's key and a **zero timestamp**. That is the account **home** document's genesis, deterministic on purpose so every device of an account bootstraps the same home document, matching the daemon's `ensureProfileGenesis`. Ed25519 signatures are deterministic, so for one signer it is always the same CID.

The agents API's `document.create` (`writeDocumentCreate`) used it for **every** document. A document's identity is its genesis, so every document an account created that way became the home document as far as the daemon is concerned. Effects: shared comment threads (the sixty conference papers with one thread), a 1970 creation date on every such document, and moves/forks indistinguishable from the home document.

| Caller | Document | Verdict |
|---|---|---|
| `client.ts publishDocument` | home only (`!input.path`) | correct |
| `create-web-universal-client.ts` | home only | correct |
| web `publishWebDocument` | home only (`publishPath.length === 0`) | correct |
| agents `publishSigningIdentityProfileAndHome` | home | correct |
| **agents `writeDocumentCreate`** | **every document** | **wrong: production offender** |
| **CLI test helper `createDocumentUpdate`** | **every new document** | **wrong: test fixtures** |
| CLI `document create` | first content Change is the genesis | correct |

## The fix

A non-home document needs no genesis blob at all: **its first content Change is its genesis**, which is what the CLI's `document create` has always done. `document.create` and the test helper now build the first Change without a `genesisCid` and point the Version Ref's `genesis` and `version` at that Change's CID. One blob fewer per document, and no timestamp ordering between a genesis and a first change to get right.

The SDK names the home helper for what it is:
- `createHomeGenesisChange(signer)`: the deterministic home genesis, with the warning and the "first change is the genesis" guidance in its doc comment.
- `createGenesisChange` stays as a **deprecated alias**, since the package is published to npm.

Every home-document caller uses the explicit name. The CLI's document command imported the helper without using it; import removed.

## Tests

- `change.test.ts`: the home genesis is stable for a signer, and the deprecated alias equals it.
[...]

PR #1132, opened 2026-09-17
fix(daemon): attribute search hits through the publishing Ref, not the genesis

juligasa:
Explore was showing papers under the wrong document, or not at all, when several docs share a genesis (republish, takeover, and the batches the old agents API put on the home genesis, see #1126). Gabo hit it on the HT'26 proceedings.

Stage 2 of `SearchEntities` re-resolved every hit by genesis and kept the most recently updated sibling. Now it resolves through the same `ref/head` edge stage 1 already used, follows move redirects, and skips hits from tombstoned generations. Stage 1 is untouched, so history search behaves the same (both versions of an edited block, unrelated edits carry the hit forward). Also dropped the Go-side moved-resource rewrite in the handler, which was sending takeover results back to the original.

Faster too. Stage 2 on a 6 GB prod desktop DB, vendored SQLite: `the` 24 ms → 3 ms, `hypertext` 26 ms → 2 ms, rare word 20 ms → 0.1 ms.

Tests: four new unit tests (shared genesis, re-created path with stale genesis, republish + takeover, food versions), plus the existing e2e search suite. Verified against a running dev daemon over gRPC.

https://claude.ai/code/session_01R4wJ413srY8C8dADmaYWaX


juligasa:
Added a second commit: an account whose name matches as both the home-document title and the Profile name now comes back once (profile row kept). Only that exact case; different names for the same account are still separate rows.

PR #1141, opened 2026-09-21
Fix desktop chrome scrolling on block restore

### LINKED ISSUE #1140: Desktop app: Severe layout issue, page is force-scrolled down
Image

BTW this is what it looks like **after refresh** and the user cant scroll up or anything to access the title bar

Reproduced in production desktop app 2026.9.9

ion-lion:
## Why now
Issue #1140 reports that refreshing a desktop document can force the whole renderer down, leaving the title bar above the viewport with no user-scrollable way back.

The route block restoration hook looks for an `overflow-y: auto|scroll` ancestor. During restoration, if the target exists before the intended document viewport is measurable, it falls back to `window.scrollTo`. Desktop wraps the document in `overflow: hidden` app chrome and disables normal body scrolling, so that fallback can move the chrome itself and strand it off-screen.

## What changed
- Stop ancestor discovery when it reaches `overflow: hidden|clip` app chrome.
- Route restoration retries while the intended inner viewport is unavailable instead of scrolling the window.
- Manual block navigation becomes a no-op in the same transient clipped state.
- Preserve mobile-web document scrolling when no clipping ancestor exists.

## Why this approach
Changing the desktop page height would treat a broader layout symptom and could alter all document sizing. Restricting block scrolling at the clipping boundary fixes the causal mechanism and retains the existing inner-scroll and mobile-window-scroll paths.

## Validation
- Exact baseline `b45c6556782bfc836f4201cf5a13204c08bc94a2`: locked regression failed because block restoration called `window.scrollTo({top: 484})` inside clipped chrome.
- Fix `9baeda13925f2af2fe4a57c6033084496ef45f99`: focused tests pass 4/4, including route restoration, manual navigation, inner scroll-container positioning, and mobile-web pinned-header behavior.
- `git diff --check` passes.
- Shared UI TypeScript was attempted but the constrained runner was OOM-killed; CI remains required.

## Follow-up
This draft still needs unchanged packaged-Electron verification against the reporter's refresh sequence before it should be marked ready.

Fixes #1140


ion-lion:
Exact-source package checkpoint for #1140:

- Exact baseline `b45c6556782bfc836f4201cf5a13204c08bc94a2` packaged successfully: https://github.com/ion-lion/seed/actions/runs/35620226094
- Exact fix `9baeda13925f2af2fe4a57c6033084496ef45f99` packaged successfully: https://github.com/ion-lion/seed/actions/runs/35621541883
- Both downloaded artifacts passed embedded source-manifest verification. SHA-256: baseline `c53f8198378d9171c29403181c6d6796f1a81796c34e54b3df65b57fb01126a4`; fix `3b11a3e42613fa86e046fb924aebd772d4d45f57491be8c3f4a13759082a16bf`.
- Exact-head PR CI is fully green, including Typecheck, Lint, shared/desktop/web units, integration, editor E2E, and aggregate checks.

I am keeping this PR in draft: the unchanged interactive Electron acceptance is not yet a valid pass/fail result. The isolated Linux runner launches the exact package and rendered onboarding once, but subsequent account-fixture interactions leave the renderer unresponsive. That runner limitation needs to be removed before claiming the reporter-sequence verification.

ion-lion:
Acceptance blocker narrowed with exact-main package evidence:

- The package and daemon boot successfully; daemon reaches ACTIVE on an isolated custom network and React starts in the expected main window.
- Copying only durable state from the older #598 fixture, removing volatile Chromium state, repairing its stale deleted-draft route, and finally reducing it to one valid published-document route still leaves the renderer unresponsive to Playwright DOM queries. The result is identical with Node and Bun Playwright clients.
- This rules out extraction/dependency failure, daemon startup, P2P isolation, the stale route, and client runtime. The reusable fixture itself is incompatible with this acceptance.

I saved the repeatable launcher and pre-DOM logs. Next is a fresh canonical fixture built directly through the packaged daemon/API, then the unchanged reload/geometry assertion on baseline and fix. PR remains draft; no verification claim yet.

ion-lion:
Pushed `6e1bc6c` to strengthen the regression boundary for the actual non-`start` fallback: a centered route block inside clipped app chrome must not call `scrollIntoView`. The previous tests already covered the `window.scrollTo` fallback and route restoration; focused coverage now passes 5/5.

I also unblocked the packaged Electron runner with a deterministic CLI-provisioned fixture and exercised exact-source baseline/fix packages. Normal published-document routes had a valid inner scroll container, so they correctly scrolled that container on both revisions and did **not** reproduce the reporter-only transient clipped-shell state. I am not claiming a false packaged before/after verdict from that non-reproducing fixture; the PR remains draft while exact-head CI reruns.

ion-lion:
# Issue 1140 transient clipped-shell packaged acceptance

Updated: 2026-09-21T19:44:00Z

The same unchanged Playwright runner and canonical clean-shutdown fixture were used against source-verified Electron packages.

- Baseline `b45c6556782bfc836f4201cf5a13204c08bc94a2`: **FAILED**. Route restoration called `window.scrollTo({top: 5849, behavior: "smooth"})` while `#scroll-page-wrapper > div` was held in the observed transient `overflow-y: hidden` state.
- Fix `9baeda13925f2af2fe4a57c6033084496ef45f99`: **PASSED**. It made zero `window.scrollTo` calls, retried across the clipped state, then scrolled the intended viewport to `5245` after it became scrollable.
- Current PR head `6e1bc6c6020f54b1511f7840816a105a0d02bd86` differs from the packaged fix only in `frontend/packages/ui/src/__tests__/use-block-scroll.test.tsx`.
- Exact-head CI: all nine checks successful; PR clean and mergeable.

The controlled transient lasts 500 ms after `#target-block` appears. This is the same state naturally observed in the fresh package (`#scroll-page-wrapper > div` initially `overflow-y: hidden`, later `scroll`) but extended deterministically so route restoration races it. The fixture is a fresh daemon/CLI-published long document with a deep route `blockRef`; the runner clones that fixture identically for baseline and fix.

See `SHA256SUMS` for runner and result hashes. Raw summaries live beside each extracted exact-source package under `transient-acceptance/summary.json`.

ion-lion:
Correction to my prior acceptance claim: the packaged test demonstrates the suspected mechanism—a `window.scrollTo({top: 5849})` call while the document viewport is transiently clipped—but it does **not** produce a screenshot of the exact user-visible failure with the desktop chrome displaced. That means #1140 is not faithfully reproduced or verified yet.

I returned this PR to draft. The focused regression and CI results remain valid, but readiness now requires visual reproduction of the reporter's refresh state (or independent confirmation that this patch fixes that exact state).

ericvicenti:
@ion-lion please update your procedures so that you don't send a PR until you have reproduced the issue.

ion-lion:
You're right. I updated my operating, pull-request, and UI-defect procedures with an explicit **no reproduction, no PR** gate:

- it applies to draft and ready PRs;
- it requires clean-baseline reproduction of the reporter's observable failure, with durable evidence, before product edits or PR creation;
- suspected mechanisms, injected intermediate states, mocked regressions, and instrumented calls without the reported observable failure do not count as reproduction;
- if reproduction is blocked, investigation stays local and I request the missing conditions on the issue instead of using a PR as a work log.

I will not reopen this PR. #1140 remains unreproduced and blocked on the exact visible refresh failure.

ion-lion:
Correction: this is a **controlled visual diagnostic, not qualifying reproduction or acceptance**.

The runner injects the tall auto-height shell that makes `window.scrollTo` visibly displace the app. Although that produces a screenshot resembling the reported symptom, instrumentation manufactured the claimed state. Under the UI reproduction gate, it cannot establish that the production refresh sequence reaches this state and is not grounds to reopen or send a PR.

Diagnostic result only:

- exact main `b45c6556`: injected shell allows hidden-root scroll to `window.scrollY=5245`, displacing desktop chrome;
- exact fix `9baeda1`: no window-scroll call and the restored inner viewport scrolls instead.

The durable record is explicitly classified as non-acceptance: hm://z6MkpVa5nMUR5ZaUEyV1SE48KTNbwTuHRd83RwLgMKc4nGU3/ion/progress/desktop-forced-scroll-1140-visual-acceptance

PR #1141 remains **closed**. #1140 remains blocked until the reporter-level refresh sequence reproduces the visible failure on a clean baseline without manufacturing it, or the missing production conditions are supplied.

